PR #422, opened 2018-09-27
[HPRO-63][HPRO-396][HPRO-403] - Order Cancel/Edit/Restore 

shyamnaru:
This PR contains Jira tickets 63, 396, and 403


views/review/orders-recent-modify.html.twig
```diff
@@ -0,0 +1,78 @@
...
+                        {% else %}
+                            <i class="fa fa-times text-danger" aria-hidden="true"></i>
+                        {% endif %}
+                    </td>
+                    <td>
+                        {{ today.displayDate(order, 'oh_created_ts') }}
+                    </td>
+                    <td align="center">
+                        {% if order.oh_type == 'edit' %}
+                            <span class="label label-success">Edited & Finalized</span>
```
stephenyeargin: [Trivial] GitHub's syntax highlighter complains that this isn't `&amp;` instead of `&`.

src/Pmi/EntityManager/EntityManager.php
```diff
@@ -51,4 +58,33 @@ public function getTimezone()
...
+        foreach ($result as $key => $value) {
+            $result[$key] = $this->parseTimestamps($value);
+        }
+        return $result;
+    }
+
+    protected function parseTimestamps(array $result)
+    {
+        foreach ($result as $key => $value) {
+            if (null !== $value && substr($key, -3, 3) == '_ts' && preg_match("/^\d{4}\-\d{2}\-\d{2}/", $value)) {
```
stephenyeargin: I can see that this is borrowed from elsewhere (and thereby already vetted for why it does this), but I feel like there are more simplified ways to validate a timestamp?

src/Pmi/Controller/ReviewController.php
```diff
@@ -29,12 +31,15 @@ class ReviewController extends AbstractController
...
-            'greatest(coalesce(created_ts, 0), coalesce(collected_ts, 0), coalesce(processed_ts, 0), coalesce(finalized_ts, 0)) AS latest_ts ' .
-            'FROM orders WHERE ' .
-            '(created_ts >= :today OR collected_ts >= :today OR processed_ts >= :today OR finalized_ts >= :today) ' .
-            'AND (site = :site OR collected_site = :site OR processed_site = :site OR finalized_site = :site) ';
-        $measurementsQuery = 'SELECT participant_id, \'measurement\' as type, id, null, created_ts, null, null, finalized_ts, null, coalesce(finalized_ts, created_ts) as latest_ts ' .
+        $ordersQuery = 'SELECT o.participant_id, \'order\' as type, o.id, o.order_id, o.created_ts, o.collected_ts, o.processed_ts, o.finalized_ts, o.finalized_samples, ' .
+            'greatest(coalesce(o.created_ts, 0), coalesce(o.collected_ts, 0), coalesce(o.processed_ts, 0), coalesce(o.finalized_ts, 0), coalesce(oh.created_ts, 0)) AS latest_ts, ' .
+            'oh.type as oh_type ' .
+            'FROM orders o ' .
+            'INNER JOIN orders_history oh ' .
```
jt2k: Shouldn't this be a left join?  Otherwise, any orders without a history item will not show up in the today view.

views/order-modify.html.twig
```diff
@@ -0,0 +1,93 @@
+{% extends 'base.html.twig' %}
+{% block title %}Order - Cancel - {% endblock %}
+{% block body %}
+    <h2 class="{% if type == 'cancel' %} text-danger {% elseif type == 'restore' %} text-success {% else %} text-primary {% endif %}">
```
jt2k: I think the default (else condition) text should not have any class (i.e. default black color).  The `text-primary` style is confusing because it appears as if the title is a link.

shyamnaru: Here the default class `text-primary` is for unlock. Yes, it's bit confusing that it looks like a link may be we should think of a different custom color for unlock? Will remove the `text-primary` for now.

jt2k: I think no custom color is fine.

views/partials/participant-orders-list.html.twig
```diff
@@ -0,0 +1,43 @@
+<div class="panel panel-default">
+    <div class="panel-heading">
+        <h3 class="panel-title"><i class="fa fa-medkit" aria-hidden="true"></i> Biobank Orders</h3>
+    </div>
+    <div class="panel-body">
+        {% for order in orders %}
+        {% if loop.index == 6 %}
```
jt2k: I like that you moved this into a partial view to reuse. When this is included from the order-modify.html.twig view, I think we want to show all orders, and not just the first 5 (otherwise, this would be confusing if the selected order is hidden). Maybe pass in an optional flag with the include to disable this div and the show all button. Also, can you fix the indentation here?

shyamnaru: Yes, I thought about it a while ago but some how missed to make that change. Will do that, thanks!

src/Pmi/EntityManager/DoctrineRepository.php
```diff
@@ -1,7 +1,7 @@
 <?php
 namespace Pmi\EntityManager;
 
-class DoctrineRepository
+class DoctrineRepository extends EntityManager
```
jt2k: I see what you're doing here, but I don't think we should be making this extend from the EntityManager class because they are not analogous.  The entity manager is a level above a specific entity repository and this inheritance can get confusing.

One way to accomplish what you have done here would be to move the timestamp parsing and timezone conversion methods to the `Util` class as static methods (and add the timezone as a method parameter).  Then, you can still have your `fetchAll` method in the entity manager, but the mismatched inheritance won't be necessary.

The `dateTimesToStrings` method is only used by the `DoctrineRepository` `insert` and `update` methods, so that can stay in the `DoctrineRepository` class.

PR #424, opened 2018-10-08
[HPRO-413] Add pmi:metrics command

stephenyeargin:
The purpose of this command is to create direct requests against the 'Metrics 2 API' rather than relying on the application code to do so.

Annotates the Metric 1 API endpoint as deprecated, cleans up a style issue.

[[HPRO-413](https://precisionmedicineinitiative.atlassian.net/browse/HPRO-413)]


src/Pmi/Console/Command/MetricsCommand.php
```diff
@@ -0,0 +1,114 @@
...
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $output->setFormatter(new OutputFormatter(true));
+
+        $start_date = $input->getOption('start_date');
+        $end_date = $input->getOption('end_date');
+        $stratification = $input->getOption('stratification');
+        $centers = join(',', $input->getOption('centers'));
+        $statuses = join(',', $input->getOption('statuses'));
+        $pretty = ($input->getOption('pretty') !== false) ? JSON_PRETTY_PRINT : false;
```
jt2k: This is really minor, but the default options parameter should be the integer `0`:  
http://php.net/manual/en/function.json-encode.php

src/Pmi/Console/Command/MetricsCommand.php
```diff
@@ -0,0 +1,114 @@
...
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+
+/**
+ * Metrics Command
+ */
+class MetricsCommand extends Command
+{
+    // Allowed stratfications in RDR
+    const STRATIFICATIONS = ['TOTAL', 'ENROLLMENT_STATUS'];
```
shyamnaru: While deploying the code we found out that one of the task in our deploy script is using php 5.5 version where array class constants are not supported so it's throwing an error.

PR #425, opened 2018-10-08
[HPRO-370][HPRO-404][HPRO-405][HPRO-406][HPRO-407] PM cancel/restore/edit

shyamnaru:


web/assets/css/app.css
```diff
@@ -310,6 +310,12 @@ table.table-samples td {
     font-size:12px;
 }
 
+.eval-revert-form {
```
jt2k: I don't have a better solution at the moment, but this (and the similar class applied to the revert form for biobank orders) should be cleaned up at some point in the future.  Relying on the absolute pixel size is a pretty brittle solution that might not render correctly across browsers/operating systems or if a user is using browser page zoom.

sql/migrations/024-alter-evaluations-table-add-column-history-id.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE evaluations
+ADD COLUMN `history_id` int DEFAULT NULL AFTER `data`;
```
jt2k: Should we add an index on this column (and on the corresponding column in the orders table)?

PR #427, opened 2018-10-10
Add modify reasons

shyamnaru:


web/assets/js/views/ModifyReasons.js
```diff
@@ -0,0 +1,22 @@
+$(document).ready(function() {
+    // Ignore non-modify reasons pages.
+    if (!$('#form_reason').length) {
+        return;
+    }
+
+    $('#form_other_text').hide();
+
+    var selected = $( "#form_reason option:selected").val();
+    var other = $("#form_reason option:contains(Other)").val();
```
jt2k: It there a reason why you are using the Other label to compare, instead of the value? Why not just compare `if (selected === 'OTHER')` since the other value should always be `OTHER`?

web/assets/js/views/ModifyReasons.js
```diff
@@ -0,0 +1,22 @@
...
+    if (other === selected) {
+        $('#form_other_text').show();
+    }
+
+    $('#form_reason').on('change', function () {
+        if ($(this).val() === other) {
```
jt2k: Comment from above applies here, too.

web/assets/js/views/ModifyReasons.js
shyamnaru: Yes, the reason for comparing this way is the form drop down values are rendered as 0,1,2,3, etc in the view.

PR #428, opened 2018-10-10
[HPRO-411] Improve error messages when auth fails

jt2k:
Previously, if the OAuth workflow or the retrieval of Google Group membership fails for some reason, a generic 500 was thrown.

This PR catches errors for both of these failures and displays a more helpful error message.

Also, by throwing the proper `Symfony\Component\Security\Core\Exception\AuthenticationException` exception, this allows the user to refresh the page to restart the auth process again, instead of just recreating the error (in the case of a failed OAuth handoff).

### DIFF
--- a/src/Pmi/Security/GoogleGroupsAuthenticator.php
+++ b/src/Pmi/Security/GoogleGroupsAuthenticator.php
@@ -4,6 +4,7 @@ namespace Pmi\Security;
 use Pmi\Application\AbstractApplication;
 use Pmi\Audit\Log;
 use Pmi\HttpClient;
+use Pmi\Util;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\Security\Core\User\UserInterface;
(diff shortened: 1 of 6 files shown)


views/error-oauth.html.twig
```diff
@@ -0,0 +1,10 @@
+{% extends 'base.html.twig' %}
+{% block title %}Error - {% endblock %}
+{% block body %}
+    <div class="alert alert-danger">
+        <p>
+            <span class="glyphicon glyphicon-thumbs-down"></span>
+            <strong>An error occurred.</strong> Sorry, there was a problem logging you in to HealthPro.  Please <a href="{{ logoutUrl }}">try again</a>.
```
shyamnaru: Getting a twig error on this line since the `logoutUrl ` is not passed in the twig params.

src/Pmi/Security/GoogleGroupsAuthenticator.php
```diff
@@ -130,7 +138,9 @@ public function onAuthenticationFailure(Request $request, AuthenticationExceptio
             } else {
                 $template = 'error-auth.html.twig';
             }
-            
+        } elseif ($exception->getMessage() === self::OAUTH_FAILURE_MESSAGE) {
+            $template = 'error-oauth.html.twig';
+            $params = [];
```
shyamnaru: Here the `logoutUrl` param is missing in the `elseif` block.

PR #433, opened 2018-10-17
[HPRO-408] Add withdrawal reason to participant page and work queue

jt2k:
[[HPRO-408](https://precisionmedicineinitiative.atlassian.net/browse/HPRO-408)]

* Adds the withdrawal reason (and justification if it exists) to the message that is displayed when viewing a withdrawn participant (screenshot below)
* Adds the withdrawal reason to the work queue (table and CSV export)
* Fixes a bug where the biospecimen and physical measurements site columns in the work queue were incorrectly associated with the default column group button.

### Example withdrawn participant error message

screen shot 2018-10-17 at 4 01 52 pm

### DIFF
--- a/src/Pmi/WorkQueue/WorkQueue.php
+++ b/src/Pmi/WorkQueue/WorkQueue.php
@@ -12,6 +12,7 @@ class WorkQueue
 
     protected $app;
 
+    // These are used to map a DataTables column index to an RDR field for sorting
     public static $wQColumns = [
         'lastName',
         'firstName',
@@ -24,6 +25,7 @@ class WorkQueue
         'consentForElectronicHealthRecordsTime',
         'consentForCABoRTime',
         'withdrawalTime',
+        'withdrawalReason',
         'recontactMethod',
         'streetAddress',
         'email',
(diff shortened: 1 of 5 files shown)


src/Pmi/WorkQueue/WorkQueue.php
```diff
@@ -271,8 +271,9 @@ public function generateTableRows($participants, $app)
             if ($participant->withdrawalStatus == 'NO_USE') {
                 $row['withdrawal'] = self::HTML_DANGER . ' <span class="text-danger">No Use</span> - ' . self::dateFromString($participant->withdrawalTime, $app->getUserTimezone());
             } else {
-                $row['withdrawal'] = ''; 
+                $row['withdrawal'] = '';
             }
+            $row['withdrawalReason'] = $e($participant->withdrawalReason);
```
shyamnaru: I think you should also add `withdrawalReason` field in the static variable `$wQColumns` after `withdrawalTime`  [field] (https://github.com/all-of-us/healthpro/blob/develop/src/Pmi/WorkQueue/WorkQueue.php#L26)

PR #434, opened 2018-10-17
[HPRO-415] Add link to open embedded PDFs

jt2k:
[[HPRO-415](https://precisionmedicineinitiative.atlassian.net/browse/HPRO-415)]

Users who's browsers are configured to not allow embedded PDFs previously had no way of downloading or opening the PDFs in a new window/tab.  This adds a link that opens the PDF in a new window.  If a browser is not configured to view PDFs at all (embedded or not), the PDF will be downloaded (and, at least in Chrome, the new tab will be automatically closed).

A direct link to PDFs has been added for biobank order labels, biobank order requisitions, and SOP documents.

### Chrome setting that breaks embedded PDFs
screen shot 2018-10-17 at 4 27 55 pm

### New open/download link
screen shot 2018-10-17 at 4 28 07 pm

### DIFF
--- a/views/help/faq.html.twig
+++ b/views/help/faq.html.twig
@@ -9,6 +9,7 @@
     <div class="page-header">
         <h2><i class="fa fa-question-circle-o" aria-hidden="true"></i> Technical FAQs</h2>
     </div>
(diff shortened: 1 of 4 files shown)


shyamnaru:
We are also embedding pdf in Technical FAQs page so I think we should also add this option over there as well.

PR #442, opened 2018-10-26
[HPRO-421] Show cross-awardee warning on all pages for a participant

shyamnaru:


src/Pmi/Application/HpoApplication.php
```diff
@@ -416,6 +416,19 @@ protected function beforeCallback(Request $request, AbstractApplication $app)
                 return $this->forwardToRoute('selectSite', $request);
             }
         }
+
+        // Display cross-awardee warning on all participant pages
+        if ($request->attributes->has('participantId')) {
+            $participantId = $request->attributes->get('participantId');
+            // Check session value
+            if (empty($this['session']->get('agreeCrossOrg_' . $participantId))) {
+                $participant = $this['pmi.drc.participants']->getById($participantId);
```
jt2k: Just wanted to confirm... the extra request here to retrieve the participant from the RDR shouldn't have any noticeable performance impact, because even though the same request will be made later on, it should be cached, right?

src/Pmi/Application/HpoApplication.php
```diff
@@ -416,6 +416,19 @@ protected function beforeCallback(Request $request, AbstractApplication $app)
...
+                // Check cross-awardee
+                if (!empty($participant) && $participant->hpoId !== $this->getSiteOrganization()) {
+                    return $this->redirectToRoute('participant', ['id' => $participantId, 'return' => $request->getRequestUri(), 'cancelType' => $request->attributes->get('_route')]);
```
jt2k: I'm hesitant to use the route name to identify where the user should be returned to if they cancel.  It's not a big deal, but these names were never meant to be user-facing.  For example, the fact that we still internally use the name "evaluation" for the PM routes could be confusing to the user (though it's very unlikely that anyone would notice and/or care).

Since we are only redirecting back to one of two places (the participant lookup page or the order lookup page), I wonder if you can remove the cancelType parameter altogether, and just use the return path to determine where to send the user.  If the return path contains the string `/order/` in the URL, we should return them to the order lookup.  This will cover the case if a user ever links directly to a particular order tab and not just the main order entry point.

src/Pmi/Application/HpoApplication.php
shyamnaru: Yes, if the cache is enabled (`rdr_disable_cache: 0`) it won't make another request.

PR #456, opened 2018-12-04
[HPRO-435] Cleanup MayoLINK config and code

jt2k:
[[HPRO-435](https://precisionmedicineinitiative.atlassian.net/browse/HPRO-435)]

Removes unused configuration items and `Pmi\Order\Mayolink\Saml` class.

Also removes the default MayoLINK endpoint and checks for complete configuration (endpoint and credentials).  The endpoint is already defined in every environment that connects MayoLINK.  (Though it wouldn't hurt if someone double-checked this).


dev_config/config.yml.dist
```diff
@@ -28,14 +28,8 @@ mysql_host: 127.0.0.1
 # Specify mock order ID to skip order creation in MayoLINK
 ml_mock_order: pmitest
 
-# Specify MayoLINK URLs to point to non-production
+# Specify MayoLINK API endpoint and credentials
 # ml_orders_endpoint: https://test.orders.mayomedicallaboratories.com
```
shyamnaru: I think we should switch this endpoint from `mayomedicallaboratories` to `mayocliniclabs`

jt2k: Good idea, and good timing. Thanks!

PR #457, opened 2018-12-10
[HPRO-414] Add support for local PHP dev server

jt2k:
This PR makes some changes to support running HealthPro locally using the built-in PHP development server, as is [recommended in the GAE PHP 7.2 migration documentation](https://cloud.google.com/appengine/docs/standard/php7/php-differences#running_your_application_locally).

* Adds a new option to the deploy script to use the PHP dev server with `bin/console pmi:deploy --local --local-php`
* Adds ability to use default session provider
* Removes use of Datastore and Memcache\* for configuration, logging, and caching when using the PHP dev server
* Rewrites site switcher to remove JS dependency and post the selected site instead of putting it in the URL (this was related to this PR because the dev server didn't route URLs with `@` characters properly).
* Uses the URL Generator for Work Queue links instead of assuming an absolute path

When running the with just `--local` (dev_appserver.py), nothing should change (except for the site switcher and work queue link generation).

\* The dashboards still use Memcache.  This PR does not fix this, so you'll likely get a Memcache error when running the local PHP dev server.


jt2k:
I forgot to mention: since GAE auth isn't available locally without using dev_appserver.py (and it is going away in the new GAE runtime), when you run `--local --local-php`, the `gae_auth` local config setting should not be set to 1.

shyamnaru:
:+1:

jt2k:
> May want to adjust [some documentation](https://github.com/all-of-us/healthpro/wiki/Developer-Onboarding) for the new run options.

@stephenyeargin Good reminder.  For now, the `--local-php` option is experimental and transitional until we fully move to the php72 runtime.

PR #461, opened 2019-01-04
[HPRO-438] Replace Retire.js with npm audit

jt2k:
[[HPRO-438](https://precisionmedicineinitiative.atlassian.net/browse/HPRO-438)]

Note that the audit output will show up twice when building, since non-local builds run `npm install` first, which will include the audit status.  `npm audit` is later run separately so that we can catch the non-0 exit code if there are vulnerabilities.

### DIFF
--- a/src/Pmi/Console/Command/DeployCommand.php
+++ b/src/Pmi/Console/Command/DeployCommand.php
@@ -156,7 +156,7 @@ class DeployCommand extends Command {
             // ensure that we are up-to-date with the latest NPM dependencies
             $output->writeln('');
             $output->writeln("Checking NPM dependencies...");
-            $this->exec("npm install");
+            $this->exec("npm install --no-audit"); // npm audit will be run below in the runJsSecurityCheck method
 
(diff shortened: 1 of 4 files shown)


stephenyeargin:
Could run `--no-audit` on the `npm install` if that annoys you.

src/Pmi/Console/Command/DeployCommand.php
```diff
@@ -561,11 +561,11 @@ private function runSecurityCheck()
 
     private function runJsSecurityCheck()
     {
-        $this->out->writeln("Running RetireJS scanner...");
-        $process = $this->exec("{$this->appDir}/node_modules/retire/bin/retire --nocache --nodepath {$this->appDir}/node_modules --jspath {$this->appDir}/web/assets/dist/js", false);
+        $this->out->writeln("Running npm audit...");
+        $process = $this->exec('npm audit', false);
         if ($process->getExitCode() == 0) {
```
shyamnaru: Can we use `===`?

PR #464, opened 2019-01-11
[HPRO-300] Switch to git-secrets for private key detection

stephenyeargin:
This switches us to use the more standardized `git-secrets` binary for scanning repositories for accidentally committed keys. The patterns are borrowed from the Workbench team.

Note that the Homebrew installed version has a bug in it that requires you to build directly from master. Per this [workaround](https://github.com/awslabs/git-secrets/issues/65#issuecomment-446519551), you can install a working version via:

```bash
brew install git-secrets --HEAD
```

[[HPRO-300](https://precisionmedicineinitiative.atlassian.net/browse/HPRO-300)]


stephenyeargin:
Note that you'll need to delete `.git/hooks/pre-commit` for the installer to run properly (will not overwrite that file if it already exists)

stephenyeargin:
This is ready to go, but we are not sending it with the next release.

PR #467, opened 2019-01-25
[HPRO-429 ] GCCL datastore integration

shyamnaru:


src/Pmi/Datastore/Entity.php
```diff
@@ -1,229 +1,50 @@
...
-    }
-    
-    public static function walkEntities(callable $callback, $sql, $params)
-    {
-        $datastore = new Datastore();
-        $datastore->walkEntities($callback, self::getSchema(), $sql, $params);
-    }
-    
-    public static function countEntities($sql, $params, $limit = 0)
+    public function gc($property, $value, $operator)
```
jt2k: This method for garbage collection is specific to the Session entity. I think it would make sense to break this up into two general-purpose methods.  One to query (with property, value, and operator) and one to delete a batch.

src/Pmi/Datastore/Entity.php
```diff
@@ -1,229 +1,50 @@
...
-        }
-        if ($gdsEntity->getKeyName()) {
-            $this->setKeyName($gdsEntity->getKeyName());
-        }
+        $datastoreClient = new DatastoreClientHelper();
+        return $datastoreClient->insert(static::getKind(), $this->data);
     }
 
-    private static function buildSqlAndParameters(array $criteria)
+    public function update($id)
```
jt2k: For both the update and delete functions, you shouldn't have to pass in an ID, since the entity already represents the correct entity. The internal properties of this object should store the necessary information.

src/Pmi/Entities/AuditLog.php
```diff
@@ -1,19 +1,12 @@
...
 class AuditLog extends Entity
 {
-    protected static function buildSchema()
+    protected static function getKind()
     {
-        return (new Schema('AuditLog'))
-            ->addString('action')
-            ->addString('user')
-            ->addString('site')
-            ->addDatetime('timestamp')
```
jt2k: If we no longer need explicitly defined schemas, how are data types handled? Is it just implicit from the data passed in for that value (i.e. by passing in a DateTime object, it automatically sets that field as a timestamp?)

src/Pmi/Entities/Configuration.php
```diff
@@ -1,14 +1,12 @@
...
 
-use GDS\Schema;
 use Pmi\Datastore\Entity;
 
 class Configuration extends Entity
 {
-    protected static function buildSchema() {
-        return (new Schema('Configuration'))
-            ->addString('key')
-            ->addString('value', false);
```
jt2k: The value property should be excluded from indexing

src/Pmi/Entities/AuditLog.php
```diff
@@ -1,19 +1,12 @@
...
-            ->addString('ip')
-            ->addString('data', false);
```
jt2k: The data property should be excluded from indexing

jt2k:
This has been merged into the `gae-php72-runtime` branch. I am going to close this PR.

PR #468, opened 2019-01-31
Dashboards > Convert progress tab to MAPI2

stephenyeargin:
While @wangy70 wraps up some work on the `RACE` stratification, I wanted to go ahead and get the base pieces up for code review.

### Background

The `/rdr/v1/Metrics` endpoint (referred internally as Metrics API 1, or MAPI1) has some legacy methods in place for counting participants and has not been updated with the most current exclusion lists. As a result, the numbers reported in the dashboard are skewed slightly upward. MAPI2 (`/rdr/v1/ParticipantCountsOverTime`) is being updated to account for these changes, but it would require an inordinate amount of effort to upgrade MAPI1 because of the way it leverages map reduce/etc. to build its counts

### What is Changing

For the user, the changes shouldn't be noticeable beyond having having more accurate counts.

On the backend, I am updating the calls so that they use MAPI2. The vocabulary for parameters and response values have also changed, which means that there was a bit refactoring in how all of that has to be handled. The `history=TRUE` flag to the API is what changes the which data set is retreived (realtime vs. historical counts).

A few related changes:

- Updated the `pmi:metrics` command to support the new flags.
- Added a `.fail()` condition to the data pulls in JavaScript so those don't spin indefinitely.


src/Pmi/Console/Command/MetricsCommand.php
```diff
@@ -66,12 +67,23 @@ protected function configure()
...
-                'Pretty formatting for JSON ouput',
-                false
+                InputOption::VALUE_NONE,
+                'Pretty formatting for JSON ouput'
+            )
+            ->addOption(
+                'debug',
+                null,
+                InputOption::VALUE_NONE,
+                'Show request debug',
```
jt2k: Remove trailing comma. Isn't this is a syntax error in PHP < 7.3? Why doesn't this throw an error for you when running using the App Engine runtime?

```suggestion
                'Show request debug'
```

stephenyeargin: Commands are run locally, so I didn't encounter that.

src/Pmi/Controller/DashboardController.php
```diff
@@ -79,36 +81,88 @@ public function metricsV2LoadAction(Application $app, Request $request)
...
+                    '56-65' => '56-65',
+                    '66-75' => '66-75',
+                    '76-85' => '76-85',
+                    '86-' => '86 and above',
+                    'UNSET' => 'UNSET'
+                ];
+                break;
+            case 'RACE':
+                $display_values = [
+                    // @TODO
```
stephenyeargin: Awaiting API work.

stephenyeargin:
All pending work is complete.

src/Pmi/Drc/RdrMetrics.php
```diff
@@ -52,4 +101,51 @@ public function metricsFields()
...
+
+        if ($num_bins == 1) {
+            array_push($date_range_bins, [$start_date, $end_date]);
+            return $date_range_bins;
+        }
+
+        $this_date = $start;
+
+        for ($i = 0; $i < $num_bins; $i++) {
+            $this_end_date = $this_date + $max_days_for_metrics_api_2;
```
jt2k: This ends up with overlapping dates at the end of one bin and the beginning of the next bin.  Which results in the merged data set containing duplicate data and counts being doubled at the bin boundaries.

screen shot 2019-02-01 at 3 51 49 pm

Also, I'm worried about iterating by a number of seconds.  In general, if you're iterating by multiples (24*60*60) you're going to run into problems with DST.  This might not have any actual negative effects, but can you make sure?  The better solution would be to either use a DateTime object as an iterator and add intervals of days.  Or use strtotime (e.g. `strtotime('+200 days', $someTimestamp)`).

stephenyeargin:
Refactored the `getDateRangeBins()` method.

PR #471, opened 2019-02-04
[DASH1-72] Track tab views in Dashboard

stephenyeargin:
Adds Google Analytics tracking to the tabs in Dashboard by setting a fake page (for now -- later version of dashboard may implement URL) and then sending a `pageview` event.

Also implements a change in the base template to mock a `ga()` object.

Reference: https://developers.google.com/analytics/devguides/collection/analyticsjs/single-page-applications

[[DASH1-72](https://precisionmedicineinitiative.atlassian.net/browse/DASH1-72)]

### DIFF
--- a/views/base-head.html.twig
+++ b/views/base-head.html.twig
@@ -47,4 +47,11 @@
     gtag('js', new Date());
     gtag('config', {{ app.getConfig('google_analytics_property')|json_encode|raw }});
 </script>
+{% else %}
+<!-- Analytics Tracking Disabled -->
+<script>
+var ga = function () {
+  console.log('Google Analytics Mock:', arguments);
+}
+</script>
 {% endif %}
--- a/views/dashboard/index.html.twig
+++ b/views/dashboard/index.html.twig
@@ -212,7 +212,7 @@
                         PLOTS_SHOWN['total-progress-nav'] = true;
                     }
                     // Send pageview event for tab
-                    ga('set', 'page', '/dasboard/total-progress');
+                    ga('set', 'page', '/dashboard/total-progress');
                     ga('send', 'pageview');
                 });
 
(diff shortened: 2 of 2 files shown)


views/base-head.html.twig
```diff
@@ -47,4 +47,11 @@
     gtag('js', new Date());
     gtag('config', {{ app.getConfig('google_analytics_property')|json_encode|raw }});
 </script>
+{% else %}
+<!-- Analytics Tracking Disabled -->
+<script>
+var ga = function () {
+  console.log('Google Analytics Mock:', arguments);
```
jt2k: TIL I can probably stop worrying about browsers that don't support `console.log`
https://developer.mozilla.org/en-US/docs/Web/API/Console/log#Browser_compatibility

views/dashboard/index.html.twig
```diff
@@ -211,6 +211,9 @@
                         renderTotalProgressPlot('spinner');
                         PLOTS_SHOWN['total-progress-nav'] = true;
                     }
+                    // Send pageview event for tab
+                    ga('set', 'page', '/dashboard/total-progress');
+                    ga('send', 'pageview');
```
jt2k: I think you can just pass the page as the third parameter

```js
ga('send', 'pageview', '/dashboard/total-progress');
```

See https://developers.google.com/analytics/devguides/collection/analyticsjs/pages#implementation

stephenyeargin: Replaced. I could see where if I were sending additional events, the `set` command might be handy.

views/base-head.html.twig
stephenyeargin: Yeah, in theory we wouldn't deploy HealthPro without the tracking code set so I didn't worry about it.

PR #475, opened 2019-02-06
[DASH1-87] Restrict realtime requests to 100 days

stephenyeargin:
The API endpoint is not fast enough to meet the request deadline of 60 seconds, even with batching. This PR enforces that limit on the frontend.

[[DASH1-87](https://precisionmedicineinitiative.atlassian.net/projects/DASH1/issues/DASH1-87)]

### DIFF
--- a/views/dashboard/index.html.twig
+++ b/views/dashboard/index.html.twig
@@ -368,6 +368,12 @@
                     $('#loading-modal').modal('show');
                   }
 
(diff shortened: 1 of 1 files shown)


stephenyeargin:
Note that testing this locally still requires selecting a single awardee, as the endpoint is already too slow to respond even on the default 7-day range.

PR #480, opened 2019-02-15
[HPRO-454] Use RDR id instead of Mayo id while retrieving an order from RDR

shyamnaru:


shyamnaru:
> Is this only an issue with mock orders in environments where orders aren't actually created in MayoLINK?

Yes, it only happens with mock orders as they have different Mayo and RDR ids.

PR #484, opened 2019-02-18
[HPRO-208] Implement system for managing and displaying page-specific notices

jt2k:
[[HPRO-208](https://precisionmedicineinitiative.atlassian.net/browse/HPRO-208)]

This adds a new table for page notices and an admin section to manage these notices.  In the application's before callback middleware, notices that apply to the current page are loaded and passed to the view for rendering.  Also, a full page option is available which will just render the base template and message without executing or rendering the requested controller action.

Screenshots of an example:

screen shot 2019-02-18 at 5 14 59 pm

screen shot 2019-02-18 at 5 15 55 pm

### DIFF
--- a/sql/healthpro/notices.sql
+++ b/sql/healthpro/notices.sql
@@ -1,0 +1,9 @@
+CREATE TABLE `notices` (
+  `id` int(11) NOT NULL AUTO_INCREMENT,
+  `url` varchar(255) NOT NULL,
+  `message` text NOT NULL,
+  `full_page` tinyint(1) NOT NULL DEFAULT 0,
+  `start_ts` timestamp NULL DEFAULT NULL,
+  `end_ts` timestamp NULL DEFAULT NULL,
+  PRIMARY KEY (`id`)
+) DEFAULT CHARSET=utf8mb4;
--- a/sql/healthpro/notices.sql
+++ b/sql/healthpro/notices.sql
@@ -5,5 +5,6 @@ CREATE TABLE `notices` (
   `full_page` tinyint(1) NOT NULL DEFAULT 0,
   `start_ts` timestamp NULL DEFAULT NULL,
   `end_ts` timestamp NULL DEFAULT NULL,
+  `status` tinyint(1) NOT NULL DEFAULT 0,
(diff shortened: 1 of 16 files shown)


views/base.html.twig
```diff
@@ -98,6 +98,13 @@
         </div>
     {% endif %}
     <div class="{% block bodycontainer %}container{% endblock %}">
+        {% if global_notices is defined and global_notices|length > 0 %}
+            {% for notice in global_notices %}
+                <div class="alert alert-warning">
```
stephenyeargin: I debated when implementing the one for the dashboard about whether we needed any other variant of this, e.g. `alert-info`. I'm guessing we would only want to use this for urgent information?

sql/healthpro/notices.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE `notices` (
+  `id` int(11) NOT NULL AUTO_INCREMENT,
+  `url` varchar(255) NOT NULL,
+  `message` text NOT NULL,
+  `full_page` tinyint(1) NOT NULL DEFAULT 0,
+  `start_ts` timestamp NULL DEFAULT NULL,
+  `end_ts` timestamp NULL DEFAULT NULL,
```
stephenyeargin: If we need to add a message of indeterminate expiration, I'm guessing the instruction will be to pick a date far enough in the future?

Side question: Do we ever need to retain these, or would we regularly clean out old messages?

views/base.html.twig
jt2k: I thought about options for how the notice is displayed (different alert classes or different presentations altogether) but decided that for now warning makes sense.  If this changes in the future, we can add some more options, but it doesn't seem necessary now.

sql/healthpro/notices.sql
jt2k: Start and end times are optional.  If you put no start or end, it will always be active as long as it is enabled.  If only a start time is entered, it will start at that time and be active indefinitely from that time on.  If only an end time is entered, it will be active immediately until the end time passes.

I imagine that we would delete notices when we are done with them.

PR #487, opened 2019-02-26
[HPRO-446] Host tutorial/training videos in HealthPro

shyamnaru:


src/Pmi/Controller/HelpController.php
```diff
@@ -227,10 +241,12 @@ public function homeAction(Application $app)
...
     }
 
-    public function videosAction(Application $app)
+    public function videosAction(Application $app, Request $request)
     {
         return $app['twig']->render('help/videos.html.twig', [
-            'videoGroups' => self::$videoGroups
+            'videoGroups' => self::$videoGroups,
+            'type' => $request->query->get('type', 'yt'),
+            'gcsBucketPath' => $app->getConfig('gcs_bucket_path')
```
jt2k: Can you add `gcs_bucket_path` to `config.yml.dist`?

shyamnaru: Sure, will do. Thanks!

jt2k: Also, can you change this to a more accurate name?  Maybe `help_videos_path` (since in theory, this wouldn't have to be in GCS)

shyamnaru: Yes, that makes sense and I also changed other variable names to be more generic.

PR #490, opened 2019-03-06
[HPRO-459] Fix order status display text

shyamnaru:


src/Pmi/Order/Order.php
```diff
@@ -1174,17 +1175,31 @@ public function createOrderHistory($type, $reason = '')
...
             $id = $ordersHistoryRepository->insert($orderHistoryData);
             $this->app->log(Log::ORDER_HISTORY_CREATE, [
                 'id' => $id,
                 'type' => $orderHistoryData['type']
             ]);
+
+            //Get most recent order edit history id if exists while restoring or reverting an order
+            if ($type === self::ORDER_REVERT || $type === self::ORDER_RESTORE) {
+                $ordersQuery = "SELECT oh.id FROM orders_history oh WHERE oh.order_id = :orderId AND oh.type = :type ORDER BY oh.id DESC LIMIT 1";
+                $orderHistory = $this->app['em']->fetchAll($ordersQuery, [
```
jt2k: Is there a reason why you can't use the `fetchBy` method on `$ordersHistoryRepository`?

shyamnaru: Didn't realize that our fetchBy has a limit option. Thanks!

PR #496, opened 2019-03-18
[DASH1-101] Dashboard > Update recruitment center GPS coordinates

stephenyeargin:
Screen Shot 2019-03-18 at 11 00 36 AM

Two of the rows in `recruitment_center_codes` have an incorrect longitude (should be negative) which causes the map view to place the points in the upper left corner. This PR introduces a migration to update those rows.

It also includes a trivial change to convert tabs to spaces in the original migration, but does not change its content.

[[DASH1-101](https://precisionmedicineinitiative.atlassian.net/projects/DASH1/issues/DASH1-101)]


shyamnaru:
I think you should also update this in dashboards sql [file](https://github.com/all-of-us/healthpro/blob/develop/sql/dashboards/recruitment_center_codes.sql)

stephenyeargin:
> I think you should also update this in dashboards sql [file](https://github.com/all-of-us/healthpro/blob/develop/sql/dashboards/recruitment_center_codes.sql)

✅  Done

shyamnaru:
Not sure if I am right but just wondering whether the dots on the middle right are misplaced as well?
Dashboard

stephenyeargin:
@shyamnaru Those are intentional. They are Direct Volunteer sites, so there isn't a geographic space to put them. At some point, the team decided to locate them off the Atlantic seaboard. 🤷‍♂️

shyamnaru:
:+1:

PR #498, opened 2019-03-20
[HPRO-462] Display cancel reason

shyamnaru:


src/Pmi/Controller/EvaluationController.php
```diff
@@ -143,6 +143,7 @@ public function evaluationAction($participantId, $evalId, Application $app, Requ
             $evaluationService->loadFromArray($evaluation);
             $evaluation['canCancel'] = $evaluationService->canCancel();
             $evaluation['canRestore'] = $evaluationService->canRestore();
+            $evaluation['reasonDisplayText'] = !empty($evaluation['eh_reason']) ? $evaluationService->getReasonDisplayText() : null;
```
jt2k: Instead of checking for an empty reason here, can you move that check into the `getReasonDisplayText` method?

src/Pmi/Evaluation/Evaluation.php
```diff
@@ -757,4 +757,15 @@ public function sendToRdr()
         }
         return false;
     }
+
+    public function getReasonDisplayText()
+    {
+        // Check only cancel reasons
+        foreach (self::$cancelReasons as $value => $key) {
```
jt2k: Can you use `array_search` here instead of looping through the array?

src/Pmi/Order/Order.php
```diff
@@ -184,6 +184,7 @@ public function loadOrder($participantId, $orderId)
         $this->participant = $participant;
         $this->version = !empty($this->order['version']) ? $this->order['version'] : 1;
         $this->order['status'] = !empty($this->order['oh_type']) ? $this->order['oh_type'] : self::ORDER_ACTIVE;
+        $this->order['reasonDisplayText'] = !empty($this->order['oh_reason']) ? $this->getReasonDisplayText() : null;
```
jt2k: Same comment as above - move the check into the `getReasonDisplayText`

src/Pmi/Order/Order.php
```diff
@@ -1504,4 +1505,15 @@ public function revertOrder($participantId)
         });
         return $status;
     }
+
+    public function getReasonDisplayText()
+    {
+        // Check only cancel reasons
+        foreach (self::$cancelReasons as $value => $key) {
```
jt2k: Same comment as above - use `array_search`

PR #499, opened 2019-03-25
[HPRO-468] Ignore full page notices for admin urls

shyamnaru:


src/Pmi/Service/NoticeService.php
```diff
@@ -38,6 +38,10 @@ public function getCurrentNotices($url)
 
         $matches = [];
         foreach ($notices as $notice) {
+            // Ignore full page notices for admin urls
+            if ($notice['full_page'] && strpos($url, 'admin') !== false ) {
```
stephenyeargin: I would guess we don't have a scenario where `admin` appears in any other part of the URL?

Example would be if we had a `/samples/admin` route, etc., this rule would activate.

jt2k: Good point. Another way to implement this is to put it in the `HpoApplication::beforeCallback` method.  You could check the current request's controller class and prevent full page notices for the `AdminController`.  This way, all actions in the Admin controller are included, and the notice still renders as a non-full page notice.

In the `beforeCallback` method, you can update the `if ($notice['full_page'])` check with:

```php
if ($notice['full_page'] && strpos($request->attributes->get('_controller'), 'Pmi\Controller\AdminController::') !== 0)
```

PR #500, opened 2019-04-05
[DASH1-85] Add EHR Metrics tab, refactor other tabs

stephenyeargin:
Screen Shot 2019-04-05 at 2 12 52 PM

This PR refactors the existing tabs to use their own controller actions so that you can link the individual tabs. It then primarily adds the new `EHRMetrics`  endpoint to build out a series of visualizations and tables.

[[DASH1-85](https://precisionmedicineinitiative.atlassian.net/browse/DASH1-85)]


stephenyeargin:
Note that the test environment does not have EHR data submitted, so those are expected to be empty at this point.

PR #504, opened 2019-04-22
[HPRO-470] Add Patient Status Flag in HealthPro (Phase 1)

shyamnaru:


views/partials/patient-status-history-dl.html.twig
```diff
@@ -0,0 +1,17 @@
...
+        <dt>User</dt>
+        <dd>{{ field.user_email }}</dd>
+        <dt>Site</dt>
+        <dd>{{ field.site_name }}</dd>
+        <dt>Date</dt>
+        <dd>{{ field.created_ts|date('n/j/Y g:ia', app.userTimezone) }}</dd>
+        <dt>Comments</dt>
+        <dd>{{ field.comments }}</dd>
+    </dl>
+    {% if key < orgPatientStatusHistoryData|length - 1 %}
```
jt2k: You can use `not loop.last` instead of having to look at the key.  [Twig documentation](https://twig.symfony.com/doc/1.x/tags/for.html#the-loop-variable)

views/participant.html.twig
```diff
@@ -457,6 +596,57 @@ $(document).ready(function() {
...
+    // Switch to default tab if empty
+    if (!orgPatientStatusData) {
+        $('[href="#patientstatus"]').tab('show');
+    }
+
+    // Switch to patient status tab if there is a form error
+    if ($('.patient-status-form').find('div').hasClass('alert-danger')) {
+        $('[href="#patientstatus"]').tab('show');
+        // Display form
+        setTimeout(function () {
```
jt2k: I remember you said you had some trouble with this. Is there a reason why you can't just call

```js
$('#patient-status-data-box').hide();
$('#patient-status-form-box').show();
```
?

views/participant.html.twig
```diff
@@ -457,6 +596,57 @@ $(document).ready(function() {
...
+        $('#patient-status-form-box').hide();
+    });
+
+    // Display history in modal window
+    var psDetailsModal = $('#patient-status-details-modal');
+    $(".patient-status-details").on('click', function (e) {
+        e.preventDefault();
+        $(psDetailsModal).removeData('bs.modal');
+        $(psDetailsModal).modal({
+            remote: $(this).attr('data-href')
```
jt2k: The remote option [is deprecated](https://getbootstrap.com/docs/3.4/javascript/#modals-options).  We can leave this for now, but we should add a task to refactor.

views/participant.html.twig
```diff
@@ -429,6 +452,122 @@
...
+                        </div>
+                    </div>
+                    {% endif %}
+                </div>
+                {% if awardeePatientStatusData is not empty %}
+                    <div class="row">
+                    <div class="col-md-12">
+                        <div class="panel panel-default">
+                            <div class="panel-heading">
+                                <h3 class="panel-title">Patient Status Details: For other organizations</h3>
```
jt2k: Can we change this to:

`Patient Status Details: Other Organizations`

views/participant.html.twig
shyamnaru: Yes, some how it didn't work when I call hide and show methods, so had to delay and trigger the click.

PR #506, opened 2019-04-23
[HPRO-475] Remove Patient Status tab and status box for DV sites

shyamnaru:


views/participant.html.twig
```diff
@@ -96,7 +96,8 @@
                 {% endif %}
                 <br />
             </div>
-            <div class="status-block patient-status-block">
+            {% if app.isDVType is empty %}
```
jt2k: Since this method queries the sites table, put it in the controller and pass to the view.

PR #512, opened 2019-05-07
[DASH1-121] Sort Table of EHR Organizations

stephenyeargin:
The values coming back from the RDR are not sorted, so this applies sorting client-side.

[[DASH1-121](https://precisionmedicineinitiative.atlassian.net/browse/DASH1-121)]


stephenyeargin:
This is rolled into #514, which remains a work in progress.

PR #515, opened 2019-05-10
[HPRO-485] Disable patient status tab (view + write) for withdrawn participants

shyamnaru:


src/Pmi/Controller/DefaultController.php
```diff
@@ -500,7 +500,11 @@ public function participantAction($id, Application $app, Request $request)
 
     public function patientStatusAction($participantId, $patientStatusId, Application $app)
     {
-        if ($app->isDVType()) {
+        $participant = $app['pmi.drc.participants']->getById($participantId);
+        if (!$participant) {
+            $app->abort(404);
+        }
+        if ($app->isDVType() || $participant->statusReason === 'withdrawal') {
```
stephenyeargin: I suppose you could combine the logic above into this one, but it may be easier to read the way you have it.

shyamnaru: Yeah, I thought of combining them initially but spitted out separately to avoid confusion.

shyamnaru: Probably, it makes sense to return 403 for non DV and withdrawn participants since that's what we are doing in other places.

PR #530, opened 2019-06-07
[DASH1-102] Add 'Participant' enrollment status

stephenyeargin:
Adds a 'Participant' status, defined as:

>A person who has completed the informed consent process and is enrolled in AoURP will become an official Participant. All Participants will have the opportunity to respond to the HIPAA Authorization/EHR Consent or the DV EHR Intent to Share agreement. If the participant agrees to the sharing EHR data, this person will move to the next status: Fully Consented. If the individual does not agree to share EHR data, the status will remain as a participant. Participants do then have the ability to complete PPI questionnaires and become eligible for PM&B, but will not be considered Core Participants.

[[DASH1-102](https://precisionmedicineinitiative.atlassian.net/browse/DASH1-102)]

PR #533, opened 2019-06-10
Revert "[DASH1-102] Add 'Participant' enrollment status"

stephenyeargin:
A timing issue between the RDR release and HealthPro means we need to hang on to this for a few more days.

Reverts all-of-us/healthpro#530

### DIFF
--- a/web/assets/js/dashboard.js
+++ b/web/assets/js/dashboard.js
@@ -92,17 +92,17 @@ function loadBarChartAnnotations(plotlyData, annotationsArray, interval) {
         fontSize = 10;
     }
     /* eslint-disable security/detect-object-injection */
-    for (var i = 0; i < plotlyData[0].x.length ; i++){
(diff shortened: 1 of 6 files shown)


web/assets/js/dashboard.js
```diff
@@ -182,22 +182,19 @@ function loadTableData(tableTarget, sourceData, colTitle) {
...
     var headerRow = $('#' + tableId + '-headers');
-    $(sourceData[0].x).each(function(i, date) {
+    $(sourceData[0]['x']).each(function(i, date) {
         headerRow.append($("<th>").text(date));
     });
 
     // load scores
     var tableBody = $('#' + tableId + '-body');
     $($(sourceData).get().reverse()).each(function(index, row) {
-        // Ignore row if no data provided
```
stephenyeargin: This was inadvertently reverted as well (fixing a bug where an RDR value that's expected (but not yet available) comes through as an empty row..

PR #532, opened 2019-06-10
[DASH1-102] Fix region map title based on selection

stephenyeargin:
I missed the code block that set the title when picking various enrollment statuses.

### Less than all selected

Screen Shot 2019-06-10 at 12 13 27 PM

### All enrollment statuses selected

Screen Shot 2019-06-10 at 12 13 35 PM

[[DASH1-102](https://precisionmedicineinitiative.atlassian.net/browse/DASH1-102)]

PR #534, opened 2019-06-10
[DASH1-102] Add 'Participant' enrollment status

stephenyeargin:
This is a resurection of #530 and #532 for a supplemental release.

[[DASH1-102](https://precisionmedicineinitiative.atlassian.net/browse/DASH1-102)]


stephenyeargin:
Also wanted to highlight the `version` parameter on this one (not to be confused with the URL segment `v1`). It is what will allow us to release this after the RDR does without breaking backwards compatibility.

PR #537, opened 2019-06-17
[HPRO-491] Use RDR patient status API endpoint to create/update patient status

shyamnaru:


src/Pmi/PatientStatus/PatientStatus.php
```diff
@@ -206,6 +222,56 @@ public function hasAccess($participant)
             !$this->app->isDVType() &&
             $participant->statusReason !== 'withdrawal' &&
             $participant->statusReason !== 'test-participant' &&
-            !$this->app->isTestSite();
+            !$this->app->isTestSite() &&
+            empty($this->app->getConfig('disable_patient_status_message'));
+    }
+
+    public function loadData($participantId, $patientStatusId, $form)
```
jt2k: Instead of passing in a Form object, I would call $form->getData() from outside this method and then pass it in.  That way, this method isn't tightly coupled with the Form object.

src/Pmi/Service/PatientStatusService.php
```diff
@@ -0,0 +1,65 @@
...
+        $this->app = $app;
+        $this->db = $app['db'];
+        $this->em = $app['em'];
+        $this->rdr = $app['pmi.drc.participants'];
+    }
+
+    public function sendPatientStatusToRdr()
+    {
+        $limit = $this->app->getConfig('patient_status_queue_limit');
+        $patientStatuses = $this->em->getRepository('patient_status')->fetchBySql("id not in (select patient_status_id from rdr_patient_status_log) limit 0, $limit");
```
jt2k: If there isn't an offset, just put the limit.  No need for `0, `

jt2k: I think I would prefer left joining to the rdr_patient_status_log and looking for nulls. Probably not a big deal either way, especially if you have a reason for doing it this way.

PR #538, opened 2019-06-19
[HPRO-474] Display patient status in WQ and WQ export

shyamnaru:


src/Pmi/WorkQueue/WorkQueue.php
```diff
@@ -18,6 +18,10 @@ class WorkQueue
         'lastName',
         'firstName',
         'dateOfBirth',
+        'patientStatus',
```
jt2k: I know these are just placeholders since sorting is disabled on these columns, but would it make sense to use the full column names anyways (patientStatusYes, patientStatusNo, etc.)

PR #539, opened 2019-06-25
[HPRO-491] Patient status optimize cron job

shyamnaru:


sql/migrations/030-create-rdr-patient-status-log-table.sql
```diff
@@ -1,6 +0,0 @@
-CREATE TABLE `rdr_patient_status_log` (
```
stephenyeargin: Could probably do this as `DROP TABLE IF EXISTS ...`, but since there are only a handful of developers and this isn't in use, it's likely fine to simply remove the migration.

src/Pmi/Service/PatientStatusService.php
```diff
@@ -22,52 +22,35 @@ public function sendPatientStatusToRdr()
...
+            } else {
+                syslog(LOG_ERR, "#{$patientStatusHistory['id']} failed sending to RDR: " . $this->rdr->getLastError());
             }
-            // Log entry in database after all the patient status history records has been successfully sent to rdr
-            $id = $this->em->getRepository('rdr_patient_status_log')->insert([
-                    'patient_status_id' => $patientStatus['id'],
-                    'created_ts' => new \DateTime()
-                ]
-            );
-            $this->app->log(Log::RDR_PATIENT_STATUS_LOG_ADD, [
```
jt2k: Remove the `Log::RDR_PATIENT_STATUS_LOG_ADD` constant, since it's not used anymore.

shyamnaru: @jt2k Thanks for catching this, I also removed table `rdr_patient_status_log` from entities.

PR #541, opened 2019-06-28
[HPRO-474] Clear patient status for withdrawn participants in WQ and Export

shyamnaru:


shyamnaru:
Initially I did this way but as it's checking for withdrawal status multiple times per each loop thought I would minimize it. But I think what you are suggesting would be clearer. Will do that.

PR #542, opened 2019-07-01
[DASH1-140] Fix filter toggles where name has parenthesis

stephenyeargin:
The prior implementation of this took the name, lowercased it, split on whitespace and then joined it back together with dashes. Trouble with that is when the name contains other characters that may trip up the selector in jQuery. This implementation abstracts that away into a custom Twig function that 'slugify's the the name a bit better, though there are still cases where an org name could trip it up. If we needed a more robust solution, I'd advise using a third-party package specialized to the task.

[[DASH1-140](https://precisionmedicineinitiative.atlassian.net/browse/DASH1-140)]

### DIFF
--- a/src/Pmi/Application/AbstractApplication.php
+++ b/src/Pmi/Application/AbstractApplication.php
@@ -373,8 +373,15 @@ abstract class AbstractApplication extends Application
         }));
 
         // Register custom Twig path_exists function
-        $this['twig']->addFunction(new Twig_SimpleFunction('path_exists', function($name) {
+        $this['twig']->addFunction(new Twig_SimpleFunction('path_exists', function ($name) {
             return !is_null($this['routes']->get($name));
+        }));
+
+        // Convert a string into a slug
+        $this['twig']->addFilter(new Twig_SimpleFilter('slugify', function ($text) {
+            $output = trim(strtolower($text));
+            $output = str_replace([' ', '-', '(', ')', ':'], '-', $output);
+            return $output;
         }));
 
         /**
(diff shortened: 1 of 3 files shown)


src/Pmi/Application/AbstractApplication.php
```diff
@@ -373,10 +373,17 @@ protected function enableTwig()
...
         // Register custom Twig path_exists function
-        $this['twig']->addFunction(new Twig_SimpleFunction('path_exists', function($name) {
+        $this['twig']->addFunction(new Twig_SimpleFunction('path_exists', function ($name) {
             return !is_null($this['routes']->get($name));
         }));
 
+        // Convert a string into a slug
+        $this['twig']->addFilter(new Twig_SimpleFilter('slugify', function ($text) {
+            $output = trim(strtolower($text));
+            $output = str_replace([' ', '-', '(', ')', ':'], '-', $output);
```
jt2k: Based on my understanding of the requirements, I'm not seeing a need to include any characters other than letters and digits.  What about something like:

```php
$output = preg_replace('/[^a-z0-9]/', '-', $output);
```

stephenyeargin: ❤️

src/Pmi/Drc/RdrHelper.php
```diff
@@ -54,10 +54,12 @@ public function getClient($resourceEndpoint = null)
             $endpoint = $this->endpoint;
         }
 
-        $client = new \Memcache();
-        $cachePool = new MemcacheCachePool($client);
+        if (class_exists('Memcache')) {
```
jt2k: Did you mean to include this change? I suppose it makes sense, but I didn't see anything in the commit message that indicates this.

stephenyeargin: No, it wasn't intentional. I needed it for debugging, and wasn't specific enough in my `git add src/Pmi` command.

PR #545, opened 2019-07-02
[HPRO-499] Work Queue export configuration

jt2k:
[[HPRO-499](https://precisionmedicineinitiative.atlassian.net/browse/HPRO-499)]

Makes the previously hard-coded export limit and page size configurable.


jt2k:
I need to make an additional change to update the export limit modal text.

shyamnaru:
Could you also add these new fields in the dist file?

PR #553, opened 2019-07-18
[DASH1-137] Add Retention PPI modules to Lifecycle dashboard

stephenyeargin:
This PR adds counts for the three retention PPI modules (as well as their aggregate full completion count) to the dashboard by way of splitting out "Baseline" and "Retention" modules into separate graphs.

- [ ] Requires related RDR change before release.

[[DASH1-137](https://precisionmedicineinitiative.atlassian.net/projects/DASH1/issues/DASH1-137)]


shyamnaru:
Noticed that the initial letters are being stripped out in the Retention PPI Module, not sure if this is an issue with the empty data?

Screen Shot 2019-08-02 at 3 12 53 PM

stephenyeargin:
Adjusted the labels so they will fit in the space.

Dashboard (2019-08-05 10-24-33)

PR #568, opened 2019-08-22
[HPRO-509] Move middle name after first name in WQ

shyamnaru:


shyamnaru:
Looks like they want to reorder in both WQ and participant search results page.

PR #575, opened 2019-10-11
[HPRO-529] Update CircleCI to use 7.2 Runtime

stephenyeargin:
This PR adjusts the configuration for CircleCI to use a PHP 7.2 Docker image to align it with an upcoming release of HealthPro.

Additionally:

- Adjusts the caching settings to preserve `vendor/` based on the hash of `composer.lock` ([Reference](https://circleci.com/docs/2.0/language-php/#sample-configuration))
- Removes several comments from the auto-generated `config.yml` file
- Drops requirements for `gcloud` to be installed, though that may come back later
- Significant performance increase to address HPRO-334.

[[HPRO-529](https://precisionmedicineinitiative.atlassian.net/browse/HPRO-529)]
[[HPRO-334](https://precisionmedicineinitiative.atlassian.net/browse/HPRO-334)]


.circleci/config.yml
```diff
@@ -1,86 +1,64 @@
...
-    - run: mkdir -p $CIRCLE_ARTIFACTS $CIRCLE_TEST_REPORTS
-    # This is based on your 1.0 configuration file or project settings
+
+    # Our primary container isn't MYSQL so run a sleep command until it's ready.
     - run:
-        working_directory: ~/vanderbilt/pmi-drc-hpo
-        command: 'sudo service mysql status || sudo service mysql start; '
+        name: Waiting for MySQL to be ready
+        command: |
+          for i in `seq 1 10`;
```
jt2k: Can we increase this? Maybe 30 seconds?

ci/test.sh
```diff
@@ -1,3 +1,5 @@
 #!/bin/bash
 
+php -v
```
jt2k: Did you mean to leave this here? It definitely doesn't hurt, but I wasn't sure if it was intended.

stephenyeargin: Going to drop it, since I believe we can be _reasonably_ sure we're going to get the right version of PHP since we're using their image. If it was more of a black-box container, it might be more useful to have this somewhere in the log.

stephenyeargin:
Ignoring Codacy review on this one as it 1) is going into another feature branch and 2) should have negligible changes affecting code quality in the scanned languages.

jt2k:
Note that there were no Codacy issues found. The problem was there was a Codacy outage that caused the check to fail.

PR #576, opened 2019-10-15
[HPRO-469] Use Symfony cache component

jt2k:
[HPRO-469]

Refactors our Datastore caching to use the Symfony cache component and a custom Datastore adapter.

[HPRO-469]: https://precisionmedicineinitiative.atlassian.net/browse/HPRO-469


src/Pmi/Controller/DashboardController.php
```diff
@@ -881,7 +881,7 @@ public function metricsLoadEHRAction(Application $app, Request $request)
      * Get Metrics 2 Object
      *
      * Main method for retrieving near-real-time metrics from API;
-     * Stores result in memcache with 15-minute expiration
+     * Caches result with 15-minute expiration
```
stephenyeargin: Noting that the cache expiration is actually handled in `RdrMetrics` instead of here. Likely an old comment before a refactoring.

src/Pmi/Drc/RdrMetrics.php
```diff
@@ -84,19 +83,28 @@ public function metrics($start_date, $end_date, $stratification, $centers, $enro
...
                 $response = $client->request('GET', 'rdr/v1/ParticipantCountsOverTime', $request_options);
                 $metrics = json_decode($response->getBody()->getContents(), true);
 
-                // Store results in Memcache if enabled
-                if ($this->memcache) {
-                    $this->memcache->set($memcacheKey, $metrics, 0, 900); // 900 s = 15 min
+                // Store results in cache if enabled
+                if ($this->cache && $metrics) {
+                    $cacheItem->set($metrics);
+                    $cacheItem->expiresAfter(900); // 15 minutes
```
stephenyeargin: Aforementioned 15-minute cache.

web/index.php
```diff
@@ -13,14 +13,10 @@
...
     $app['sessionTimeout'] = 3600 * 24; // Extend session time out in local environment
-    if (!$app->isPhpDevServer()) {
-        $app['twigCacheHandler'] = 'memcache';
-        $app['sessionHandler'] = 'datastore'; // use datastore for dev_appserver local environment
-    }
 } else {
     $app['sessionHandler'] = 'datastore';
-    $app['cacheDirectory'] = sys_get_temp_dir();
     $app['twigCacheHandler'] = 'file';
+    $app['twigCacheDirectory'] = realpath(__DIR__ . '/../cache') . '/twig';
```
stephenyeargin: Noting that this revert is because the Twig files are compiled/cached locally and then sent up as part of the deployment process. We're not actually ever attempting to write to this folder in a running app (which wouldn't be allowed).

PR #579, opened 2019-10-18
[HPRO-534] Docker containers for HealthPro

stephenyeargin:
In an effort to standardize our development environment as we move to the new App Engine run-time, we will create a set Docker containers that replicate (approximately) the environments that we use in production. This approach should reduce the amount of effort needed to onboard new developers, and lay the groundwork for improvements in our local testing process.

[HPRO-534]

[HPRO-534]: https://precisionmedicineinitiative.atlassian.net/browse/HPRO-534


src/Pmi/Console/Command/DeployCommand.php
```diff
@@ -153,7 +153,7 @@ protected function execute(InputInterface $input, OutputInterface $output)
         }
 
         if ($this->local) {
-            $cmd = "php -S localhost:{$this->port} -t web/ web/local-router.php";
+            $cmd = "php -S 0.0.0.0:{$this->port} -t web/ web/local-router.php";
```
stephenyeargin: Docker's port forwarding seems to require the server not be bound to `localhost` if you want it to be available on the host machine.

jt2k:
Additional known issue: need to start up datastore emulator

stephenyeargin:
Remaining task status:

###Bootstrapping the database schema on build

Haven't found a way to do this beyond:

```
docker exec -it healthpro-web docker/database.sh
```

### Documenting its use in the README/Wiki

Should we plan to update the [Developer Onboarding](https://github.com/all-of-us/healthpro/wiki/Developer-Onboarding) document instead? New doc?

### Aligning versions with what's available in GAE

Some of the lower-level stuff may not be locked at a particular version. We're already matching PHP and MySQL.

PR #583, opened 2019-10-29
[HPRO-538 ] Fix empty withdrawal timestamp issue in participant details page

shyamnaru:


views/participant.html.twig
```diff
@@ -614,7 +616,9 @@
                                         <div class="panel-body">
                                         <p>{{ errorMessage }}</p>
                                         {% if participant.statusReason == 'withdrawal' %}
-                                            <p><strong>Withdrawal Time:</strong> {{ participant.withdrawalAuthored|date('n/j/Y g:ia', app.userTimezone) }}</p>
+                                            {% if participant.withdrawalAuthored is not empty %}
+                                                <p><strong>Withdrawal Time:</strong> {{ participant.withdrawalAuthored ? participant.withdrawalAuthored|date('n/j/Y g:ia', app.userTimezone) }}</p>
```
jt2k: I didn't realize we have this block duplicated from above. For now, since it is duplicated, let's at least keep it the same. This line adds an extra empty check that isn't in the first block. It's redundant since this whole line is wrapped in the if condition, so I think using the same line from above here makes sense.

shyamnaru: I thought of implementing this way initially and then forgot to remove this piece of code.

PR #584, opened 2019-11-06
[HPRO-540] Allow defining DB connection in docker-compose configuration

jt2k:
HPRO-540

This change allows for environment variables to override the DB configuration when running locally and sets the necessary environment variables in the `healthpro-web` container.

As a result of this change, you will no longer need to change the local `dev_config/config.yml` file to match your Docker configuration.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,7 +2,7 @@ version: '2'
 services:
     mysql:
         environment:
-            MYSQL_ROOT_PASSWORD: ~
+            MYSQL_ROOT_PASSWORD: ''
             MYSQL_ALLOW_EMPTY_PASSWORD: 'yes'
             MYSQL_DATABASE: hpo
         container_name: healthpro-mysql
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -14,6 +14,11 @@ services:
             - ./sql/healthpro:/docker-entrypoint-initdb.d
 
     web:
+        environment:
+            MYSQL_HOST: healthpro-mysql
+            MYSQL_DATABASE: hpo
+            MYSQL_USER: root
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -14,6 +14,11 @@ services:
             - ./sql/healthpro:/docker-entrypoint-initdb.d
 
     web:
+        environment:
+            MYSQL_HOST: healthpro-mysql
+            MYSQL_DATABASE: hpo
+            MYSQL_USER: root
+            MYSQL_PASSWORD: ''
```
shyamnaru: May be we should have a override file `docker-compose.override.yml` to set our local db configurations and include it in `.gitignore` file so that they won't get git tracked when we change anything.

I tried setting `docker-compose.override.yml` and it worked!

jt2k: My thought was that since these are already defined in this file for the mysql container definition, it made sense to include here as well. Since these aren't sensitive credentials, this allows it to work out of the box without any additional steps.

stephenyeargin: I like the idea of including `docker-compose.override.yml` in `.gitignore` at any rate in case we'd like to test settings (or switch other settings to test). But yeah, "works out of the box" is a pretty good goal as well.

jt2k: @stephenyeargin Shyam and I were discussing over Slack. It turns out he was having a problem with the two settings not working. Can we change the mysql password to `''` instead of `~`. I think there might be something unexpected happening when it's null. Does that make sense to you?

stephenyeargin: Makes sense to me. I guess the behavior of `MYSQL_ALLOW_EMPTY_PASSWORD` means "can be set to an empty string" and not `NULL`.

PR #586, opened 2019-11-07
[HPRO-512] Validate biobank ID in participant lookup

shyamnaru:


src/Pmi/Controller/BiobankController.php
```diff
@@ -33,15 +33,28 @@ public function homeAction(Application $app)
...
                     new Constraints\NotBlank(),
-                    new Constraints\Type('string')
+                    new Constraints\Type('string'),
+                    new Constraints\Callback(function($bioBankId, $context) use ($bioBankIdPrefix) {
+                        if (empty($bioBankIdPrefix)) {
+                            return;
+                        }
+                        $bioBankId = trim($bioBankId);
+                        $bioBankIdPrefix = preg_quote($bioBankIdPrefix, '/');
+                        if (!preg_match("/^{$bioBankIdPrefix}\d*$/", $bioBankId)) {
```
stephenyeargin: Trivial, but wouldn't it `\d+` instead of `\d*` given that not providing a digit is invalid.

src/Pmi/Controller/BiobankController.php
```diff
@@ -33,15 +33,28 @@ public function homeAction(Application $app)
...
     {
+        $bioBankIdPrefix = $app->getConfig('biobank_id_prefix');
         $idForm = $app['form.factory']->createNamedBuilder('id', Type\FormType::class)
             ->add('biobankId', Type\TextType::class, [
                 'label' => 'Biobank ID',
                 'constraints' => [
                     new Constraints\NotBlank(),
-                    new Constraints\Type('string')
+                    new Constraints\Type('string'),
+                    new Constraints\Callback(function($bioBankId, $context) use ($bioBankIdPrefix) {
```
jt2k: Instead of a complex callback constraint, I think it would be more straightforward to use a [Regex](https://symfony.com/doc/current/reference/constraints/Regex.html) constraint. You can see an example usage of this in the `OrderController` for validating kit ids. The only downside is that you will need to build the contraints array first and dynamically add the regex constraint if the prefix is configured, but I think the benefits of using a simple constraint are worth it.

src/Pmi/Controller/BiobankController.php
```diff
@@ -33,15 +33,24 @@ public function homeAction(Application $app)
...
+        $bioBankIdPrefix = $app->getConfig('biobank_id_prefix');
+        $constraints = [
+            new Constraints\NotBlank(),
+            new Constraints\Type('string')
+        ];
+        if (!empty($bioBankIdPrefix)) {
+            $bioBankIdPrefix = preg_quote($bioBankIdPrefix, '/');
+            $constraints[] = new Constraints\Regex([
+                'pattern' => "/^{$bioBankIdPrefix}\d+$/",
+                'message' => "Invalid biobank ID. Must be in the format of {$bioBankIdPrefix}000000000"
```
jt2k: Sorry, one more detail - the prefix in the validation message should not be preg_quoted.

PR #590, opened 2019-11-15
[HPRO-542] Add warning to the "Process" tab in the biospecimen collection page

shyamnaru:


views/order-process.html.twig
```diff
@@ -135,6 +135,20 @@ $(document).ready(function () {
     var orderView = new PMI.views['OrderSubPage']({
         el: $("body")
     });
+
+    $('.process-form button[type="submit"]').on('click', function () {
```
stephenyeargin: May not matter in a typical workflow, but what if a user presses enter while in a field to submit the form? (We may have that blocked elsewhere.)

shyamnaru: We have a `disable-enter` class in the form, which prevents triggering enter.

PR #594, opened 2019-12-06
[HPRO-550] Fix cron job cache keys delete errors

shyamnaru:


src/Pmi/Datastore/DatastoreClientHelper.php
```diff
@@ -22,9 +22,12 @@ public function __construct()
         ]);
     }
 
-    public function fetchAll($kind)
+    public function fetchAll($kind, $limit = null)
     {
         $query = $this->datastore->query()->kind($kind);
+        if ($limit) {
+            $query = $query->limit($limit);
```
jt2k: The Query methods mutate the object, so you don't need to assign it back to $query. The methods return itself for chaining purposes. In this case, you just need to call `$query->limit($limit);`

src/Pmi/Datastore/DatastoreClientHelper.php
```diff
@@ -62,11 +65,14 @@ public function deleteBatch($keys)
...
     }
 
-    public function basicQuery($kind, $property, $value, $operator)
+    public function basicQuery($kind, $property, $value, $operator, $limit)
     {
         $query = $this->datastore->query()
             ->kind($kind)
             ->filter($property, $operator, $value);
+        if ($limit) {
+            $query = $query->limit($limit);
```
jt2k: Same comment as above - no need to assign to $query.

PR #599, opened 2019-12-10
[DASH1-173][HPRO-556] Rename 'Fully Consented' to 'Participant + EHR Consent'

stephenyeargin:
Note that this PR, in its current form, is a global find/replace for the label, so it touches Work Queue. We may want to limit it only to the Dashboard component for now.

[DASH1-173]
[[HPRO-556](https://precisionmedicineinitiative.atlassian.net/browse/HPRO-556)]


views/dashboard/partials/modal-dashboard-help.html.twig
```diff
@@ -83,10 +83,10 @@
...
+                <strong>Participant + EHR Consent (Participant + EHR):</strong> A participant with EHR consent has completed the primary informed consent process and the HIPAA Authorization/EHR consent (if available and where relevant, based on AoURP capacity to effectively search for and access records). This definition takes into account DV EHR sharing. A Participant with EHR Consent may complete research program activities such as completion of questionnaires or providing physical measurements and biospecimens, but will remain in the Participant+EHR Consent status until all initial steps of the program are completed, qualifying the participant for Core Participant status.
```
shyamnaru: Do we still require text in parentheses?

stephenyeargin: I wouldn't think so. I'll remove them. :+1:

PR #602, opened 2019-12-11
[HPRO-555] Change id columns for awardees and organizations table to VARBINARY

jt2k:
HPRO-555

[MySQL VARCHAR columns ignore trailing whitespace](https://dev.mysql.com/doc/refman/5.7/en/char.html). This led to an issue in the test environment where the test awardee API returned two organizations with the same id but one with a trailing space, resulting in a duplicate key collision. This change treats the id as a binary column to avoid the MySQL VARCHAR padding behavior.

The side affect of this is that comparisons are now case sensitive, but the IDs should be case sensitive anyways. Also, sorting will be affected, but we are not sorting organizations or awardees by ID anywhere.

There are places where we join to the awardees.id or organizations.id column from columns that are still VARCHAR. We should determine whether we want to make this change across all tables.

### DIFF
--- a/sql/healthpro/awardees.sql
+++ b/sql/healthpro/awardees.sql
@@ -1,5 +1,5 @@
 CREATE TABLE `awardees` (
-  `id` varchar(20) NOT NULL,
+  `id` varbinary(255) NOT NULL,
   `name` varchar(255) NOT NULL,
   PRIMARY KEY (`id`)
 ) DEFAULT CHARSET=utf8mb4;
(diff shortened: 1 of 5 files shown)


stephenyeargin:
Is there any need to fix existing data, or has that already been handled?

jt2k:
@stephenyeargin Not in HealthPro. I am unclear on the state of things in the upstream systems, but our sync will handle any changes that are made.

PR #604, opened 2019-12-30
[DASH1-168] Add Participant Origin to Tab 1, 3 and 4

stephenyeargin:
Allows the user to filter results based on which portal brought in the participant.

- [ ] Requires DA-1370

[DASH1-168]


stephenyeargin:
Worked out an issue where we can include the 'Participants by Region' tab (3) as well.

PR #605, opened 2020-01-09
[HPRO-557] Add additional request information to custom Stackdriver logs

shyamnaru:


src/Pmi/Application/AbstractApplication.php
```diff
@@ -577,4 +585,44 @@ public function registerCache()
         $this['cache'] = new \Pmi\Cache\DatastoreAdapter($this->getConfig('ds_clean_up_limit'));
         $this['cache']->setLogger($this['logger']);
     }
+
+    public function getSiteMetaData()
```
jt2k: I think `getLogMetadata` would be a more accurate name for this method

src/Pmi/Audit/Log.php
```diff
@@ -62,36 +62,10 @@ protected function buildLogArray()
...
-            Request::setTrustedProxies($originalTrustedProxies);
-
-            // identify cron user
-            if ($logArray['user'] === null && $request->headers->get('X-Appengine-Cron') === 'true') {
-                $logArray['user'] = 'Appengine-Cron';
-            }
-        } else {
-            $logArray['ip'] = null;
-        }
+        $siteMetaData = $this->app->getSiteMetaData();
```
jt2k: You can simplify this by doing:
```php
$logArray = array_merge($logArray, $this->app->getLogMetadata());
```
(Using renamed method as suggested above)

PR #611, opened 2020-01-29
[HPRO-561] Add deactivation details to work queue and participant summary page

shyamnaru:


src/Pmi/Entities/Participant.php
```diff
@@ -116,6 +117,12 @@ public function parseRdrParticipant($participant)
         if (!empty($participant->enrollmentStatus) && $participant->enrollmentStatus === 'FULL_PARTICIPANT') {
             $this->isCoreParticipant = true;
         }
+
+        // Set activity status
+        $this->activityStatus = $this->getActivityStatus($participant);
```
jt2k: I can't remember why we/I used this pattern of storing and accessing computed values in public variables. I don't think we should clean this up in this ticket/PR, but we should revisit at some point.

src/Pmi/WorkQueue/WorkQueue.php
```diff
@@ -9,6 +9,7 @@ class WorkQueue
     const HTML_SUCCESS = '<i class="fa fa-check text-success" aria-hidden="true"></i>';
     const HTML_DANGER = '<i class="fa fa-times text-danger" aria-hidden="true"></i>';
     const HTML_WARNING = '<i class="fa fa-question text-warning" aria-hidden="true"></i>';
+    const HTML_NOTICE = '<i class="fa fa-exclamation-circle text-warning" aria-hidden="true"></i>';
```
jt2k: The ticket says to use an "open circle". I suggest that we use `fa-stop-circle` (I think the `fa-circle-o` icon is too thin).

src/Pmi/Controller/WorkQueueController.php
```diff
@@ -425,6 +436,8 @@ public function exportAction(Application $app, Request $request)
                         $row[] = $participant->enrollmentStatusCoreStoredSampleTime;
                     }
                     $row[] = $participant->participantOrigin;
+                    $row[] = $participant->isSuspended ? '1' : '0';
```
jt2k: Ticket says to only add these columns for full-data access export.

PR #615, opened 2020-02-12
[DASH1-176] Participant Origination Filtering for Tab 2

stephenyeargin:
Screen Shot 2020-02-12 at 2 14 50 PM

Enable filtering by Participant Origination in Tab 2 (Real-Time Data).

[DASH1-176]


jt2k:
@stephenyeargin I added the `rdr dependency` label because I thought we weren't ready for this to go out with the next release until we confirmed when the RDR's change that supports this is ready for stable/prod. Did I misunderstand our conversation earlier today?

stephenyeargin:
Missed the label on mobile, and misread the version in the releases channel. It's going with v64. I believe we should still be on track for the next cycle though.

PR #617, opened 2020-02-19
[HPRO-572] Update the language for the FISMA banner in HealthPro

shyamnaru:


views/templates/system-usage.html.twig
```diff
@@ -2,7 +2,7 @@
     <div class="alert alert-warning">
         <span class="glyphicon glyphicon-warning-sign"></span>
         <strong>Warning Notice</strong>
```
stephenyeargin: It's probably not essential for this bit of legal jargon, but the style guide usually calls for "_All of Us_" to be in italics (`<em>`).

PR #618, opened 2020-02-19
[HPRO-523] Add local auth option

shyamnaru:


views/error-gae-auth.html.twig
```diff
@@ -3,6 +3,6 @@
 {% block body %}
     <div class="alert alert-danger">
         <span class="glyphicon glyphicon-thumbs-down"></span>
-        <strong>Access denied!</strong> You are not logged into Google, please <a href="{{ loginUrl }}">try again</a>.
+        <strong>Access denied!</strong> You are not logged in, please <a href="{{ path('mockLogin') }}">Login</a>.
```
jt2k: Since the link is part of a sentence ("please login"), I suggest changing the case:

```suggestion
        <strong>Access denied!</strong> You are not logged in, please <a href="{{ path('mockLogin') }}">login</a>.
```

src/Pmi/Controller/DefaultController.php
```diff
@@ -571,4 +572,34 @@ public function hideTZWarningAction(Application $app, Request $request)
         $request->getSession()->set('hideTZWarning', true);
         return (new JsonResponse())->setData([]);
     }
+
+    public function mockLoginAction(Application $app, Request $request)
+    {
+        if (!$app->isLocal() && !$app->getConfig('gae_auth')){
+            return $app->abort(403);
+        }
+        $loginForm = $app['form.factory']->createNamedBuilder('login', FormType::class)
```
jt2k: Can we prefill this with a value (`test@example.com` is fine), instead of setting the placeholder? That way we can just click the submit button.

src/Pmi/Application/AbstractApplication.php
```diff
@@ -294,29 +306,21 @@ public function getGoogleLogoutUrl($route = 'home')
...
-    }
-
     public function getUser()
     {
         $token = $this['security.token_storage']->getToken();
-        return $token ? $token->getUser() : null;
+        if ($token) {
+            return is_object($token->getUser()) ? $token->getUser() : null;
+        }
+        return null;
```
jt2k: I think this is a bit easier to read:

```suggestion
        if ($token && is_object($token->getUser())) {
            return $token->getUser();
        } else {
            return null;
        }
```

src/Pmi/Application/AbstractApplication.php
```diff
@@ -277,14 +275,28 @@ public function setConfig($key, $val)
...
+            $this['session']->set('mockUser', $cls::getCurrentUser());
+        }
     }
 
     public function getGoogleUser()
     {
-        if ($this->getConfig('gae_auth')) {
-            $cls = $this->getGoogleServiceClass();
-            return class_exists($cls) ? $cls::getCurrentUser() : null;
+        if ($this->isLocal() && $this->getConfig('gae_auth')) {
```
jt2k: Can we change this config name from `gae_auth` to something like `local_mock_auth`, since we are no longer using the old GAE auth workflow.

src/Pmi/Controller/DefaultController.php
```diff
@@ -571,4 +572,34 @@ public function hideTZWarningAction(Application $app, Request $request)
         $request->getSession()->set('hideTZWarning', true);
         return (new JsonResponse())->setData([]);
     }
+
+    public function mockLoginAction(Application $app, Request $request)
+    {
+        if (!$app->isLocal() && !$app->getConfig('local_mock_auth')){
```
jt2k: This should be an OR

src/Pmi/Security/GoogleGroupsAuthenticator.php
```diff
@@ -163,7 +162,7 @@ public function onAuthenticationSuccess(Request $request, TokenInterface $token,
         // has the user agreed to the system usage agreement this session?
         $request->getSession()->set('isUsageAgreed', false);
         
-        if ($this->app->getConfig('gae_auth')) {
+        if ($this->app->getConfig('local_mock_auth')) {
```
jt2k: This should also check for isLocal. In most places, you check for both the config setting and the local environment, but here and a few other places, you are only checking for the config setting. You might consider adding a helper app method to combine these checks.

src/Pmi/Security/GoogleGroupsAuthenticator.php
```diff
@@ -180,7 +179,7 @@ public function supportsRememberMe()
     
     public function start(Request $request, AuthenticationException $authException = null)
     {
-        if ($this->app->getConfig('gae_auth')) {
+        if ($this->app->getConfig('local_mock_auth')) {
```
jt2k: See above

PR #619, opened 2020-02-26
[HPRO-575] [HPRO-576] Send deactivate email notifications and include participant ids in emails

shyamnaru:
This PR contains JIRA tickets 575 and 576


sql/healthpro/deactivate_log.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TABLE `deactivate_log` (
+  `id` int(11) NOT NULL AUTO_INCREMENT,
+  `participant_id` varchar(50) NOT NULL,
+  `insert_ts` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  `deactivate_ts` timestamp NULL DEFAULT NULL,
+  `hpo_id` varchar(100) DEFAULT NULL,
+  `email_notified` varchar(2000) DEFAULT NULL,
+  PRIMARY KEY (`id`),
+  UNIQUE KEY `participant_id` (`participant_id`),
+  UNIQUE KEY `deactivate_ts` (`deactivate_ts`),
```
jt2k: This should be a single unique key on both the `participant_id` and `deactivate_ts` columns

shyamnaru: Initially, I tried this and then modified the schema directly in my local db so it worked while trying to insert same `participant_id` and different `deactivate_ts` and vice versa,  but forgot to update here. Will update that. Thanks!

shyamnaru: Initially, I tried this and then modified the schema directly in my local db so it worked while trying to insert same `participant_id` and different `deactivate_ts` and vice versa,  but forgot to update here. Will update that. Thanks!

PR #622, opened 2020-03-02
[HPRO-582] Update Docker configuration to use GAE image

stephenyeargin:
*Note:* @jt2k points out that this will put on the Google App Engine *Flex* environment image, rather than *Standard*. So the actual benefit here may end up somewhat moot. Still, we're moving closer to the versions used in _one_ of the environments available rather than guessing.

---

This appears to run without issue. I left the install steps in place for the emulator, `git`, etc., but we could try removing those one-by-one during more comprehensive testing. Right now, they aren't hurting anything to be side loaded.

[HPRO-582]

[HPRO-582]: https://precisionmedicineinitiative.atlassian.net/browse/HPRO-582

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM php:7.2
+FROM gcr.io/gcp-runtimes/php72:2020-02-19-11-00
 
 # Fix for issue with OpenJDK install
 RUN mkdir -p /usr/share/man/man1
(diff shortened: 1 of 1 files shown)


stephenyeargin:
Confirmed that the `gcloud` SDK is not installed by default on the image. As best I can tell, everything else we install/load is necessary.

```bash
$ docker run -it gcr.io/gcp-runtimes/php72:2020-02-19-11-00 bash
root@9875135223b9:/app# which gcloud 
root@9875135223b9:/app# 
```

stephenyeargin:
We'll revisit this at some point in the future.

PR #630, opened 2020-03-17
[HPRO-591] Make scripps/biobank order ready only view available to scripps super user

shyamnaru:
I'm not sure if want to display participant id in the order detail page for scripps super user. We can display participant id by checking roles but when a user has both (ROLE_AWARDEE and ROLE_BIOBANK) roles the biobank non-pii view will display the participant id.


jt2k:
@shyamnaru Regarding your comment about the participant ID, I think it makes sense that any user who has the `ROLE_AWARDEE` role would still see the participant ID.

PR #634, opened 2020-03-26
[HPRO-569] Display information related to Genetic Return of Results (gROR) consent

shyamnaru:


views/partials/participant-notice.html.twig
```diff
@@ -10,6 +10,12 @@
 {% elseif participant.statusReason == 'withdrawal' %}
     {% set errorTitle = 'Withdrawn participant' %}
     {% set errorMessage = 'This participant has withdrawn from the All of Us Research Program. No new data can be collected or sent for this participant. Staff must also follow all site-specific SOPs to ensure all All of Us Research Program data are removed from all local systems for this participant.' %}
+{% elseif participant.statusReason == 'genomics' %}
+    {% set errorTitle = 'Genomic Return of Results Consent not complete' %}
+    {% set errorMessage = 'This participant has provided general consent but has not completed Genomic Return of Results consent in the Participant Portal and is therefore ineligible to be scheduled for or complete physical measurements and biospecimen collection. Please direct the participant to complete the Genomic Return of Results consent in the Participant Portal to enable in-person enrollment or contact the Help Desk, if needed.' %}
+{% elseif participant.statusReason == 'ehr-consent' %}
+    {% set errorTitle = 'The EHR consent not complete' %}
```
stephenyeargin: To be consistent with the error above, perhaps this should say:

>EHR Consent not complete

src/Pmi/Controller/WorkQueueController.php
```diff
@@ -356,6 +357,8 @@ public function exportAction(Application $app, Request $request)
             if ($hasFullDataAccess) {
                 $headers[] = 'Deactivation Status';
                 $headers[] = 'Deactivation Date';
+                $headers[] = 'gROR Consent Status';
```
stephenyeargin: Should it have a lowercase `o` (e.g. `gRoR`) for consistency?

sql/healthpro/salivary_inclusion_zip_codes.sql
```diff
@@ -0,0 +1,4 @@
+CREATE TABLE `salivary_inclusion_zip_codes` (
```
stephenyeargin: We have a migration for the data for this table (`sql/migrations/033-insert-salivary-inclusion-zip-codes.sql`), but nothing that would bootstrap it for a new development environment or unit tests (two cases where we usually don't run the migrations). Not sure if this has to be addressed as part of this PR, but it may call for us to handle that differently in the future if this is a critical data fixture.

src/Pmi/Controller/WorkQueueController.php
shyamnaru: There are some inconsistencies in the ticket itself. I'll check with bhinnata and finalize one. Thanks!

sql/healthpro/salivary_inclusion_zip_codes.sql
shyamnaru: Decided to remove these migration files completely as we no longer needed them. Thanks!

src/Pmi/Entities/Participant.php
```diff
@@ -69,6 +73,15 @@ public function parseRdrParticipant($participant)
             $this->status = false;
             $this->statusReason = 'withdrawal';
         }
+        if (!empty($this->genomicsStartTime) && isset($participant->consentForStudyEnrollmentAuthored) && $participant->consentForStudyEnrollmentAuthored > $this->genomicsStartTime) {
```
jt2k, line 76: Unlikely that there will be a consent time at the exact same time as the start time, but the ticket says this should be `>=`

```suggestion
        if (!empty($this->genomicsStartTime) && isset($participant->consentForStudyEnrollmentAuthored) && $participant->consentForStudyEnrollmentAuthored >= $this->genomicsStartTime) {
```

PR #635, opened 2020-03-26
[HPRO-588] Migrate symfony components to v4.4

shyamnaru:


jt2k:
@shyamnaru This is looking good to me. We did decide that we don't want to merge this into develop, right? Can you update the destination branch for this PR?

PR #637, opened 2020-03-31
[HPRO-596] Load php.ini file in Docker image for development

stephenyeargin:
Maps a project directory into PHP's configuration directory so we can load configuration for the Docker image.

[HPRO-596]

[HPRO-596]: https://precisionmedicineinitiative.atlassian.net/browse/HPRO-596

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,7 +9,6 @@ php.ini
 /cache/
 /dev_config/*
 !dev_config/config.yml.dist
-!dev_config/php.conf.d/*
 .idea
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -9,6 +9,7 @@ php.ini
 /cache/
 /dev_config/*
 !dev_config/config.yml.dist
+!dev_config/php.conf.d/*
```
jt2k, line 12: This line can be removed now
```suggestion
```

PR #638, opened 2020-04-02
[HPRO-589] Add 'Today's Quest Biobank Orders' Tab

stephenyeargin:
Screen Shot 2020-04-08 at 4 12 45 PM

Add a view of orders matching the current date from the Quest Biobank data source. Non-production deployments can use the `?days=<int>` flag for testing purposes.

Also centralized the tabs into a partial to reduce repeated code.

[HPRO-589]

[HPRO-589]: https://precisionmedicineinitiative.atlassian.net/browse/HPRO-589

### DIFF
--- a/src/Pmi/Controller/BiobankController.php
+++ b/src/Pmi/Controller/BiobankController.php
@@ -268,11 +268,9 @@ class BiobankController extends AbstractController
         $today = $startTime->format('Y-m-d');
 
         $quanumOrders = $app['pmi.drc.participants']->getOrders([
-            'participant_id' => $participant_id,
             'startDate' => $today,
             'endDate' => date('Y-m-d'),
             'origin' => 'careevolution',
-            'kitId' => null,
             'page' => '1',
             'pageSize' => '1000'
         ]);
--- a/src/Pmi/Controller/BiobankController.php
+++ b/src/Pmi/Controller/BiobankController.php
@@ -22,6 +22,7 @@ class BiobankController extends AbstractController
         ['order', '/{biobankId}/order/{orderId}'],
         ['quanumOrder', '/{biobankId}/quanum-order/{orderId}'],
         ['ordersToday', '/review/orders/today'],
+        ['quanumOrdersToday', '/review/quanum-orders/today'],
         ['ordersUnfinalized', '/review/orders/unfinalized'],
         ['ordersUnlocked', '/review/orders/unlocked'],
         ['ordersRecentModify', '/review/orders/recent/modify']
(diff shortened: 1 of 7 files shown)


src/Pmi/Order/Order.php
```diff
@@ -1396,13 +1396,16 @@ public function loadFromJsonObject($object)
...
         // Extract participantId
         preg_match('/^Patient\/(P\d+)$/i', $object->subject, $subject_matches);
+        $participantId = $subject_matches[1];
 
         $this->order['id'] = $object->id;
-        $this->order['participant_id'] = $subject_matches[1] ? $subject_matches[1] : 'Unknown';
+        $this->order['participant_id'] = $participantId;
         $this->order['order_id'] = $kitId;
         $this->order['rdr_id'] = $object->id;
+        $this->order['biobank_id'] = $object->biobankId;
```
shyamnaru, line 1405: Currently, the RDR doesn't return biobankId right? Is this for the future behavior?

stephenyeargin: It's now returning it in test as of yesterday afternoon.

https://github.com/all-of-us/raw-data-repository/pull/1674
https://github.com/all-of-us/raw-data-repository/pull/1675

shyamnaru: I'm getting `Undefined property: stdClass::$biobankId` when I tried to view some kit orders(Y822546576, Y162243428, Y215584195) in test

stephenyeargin: I see what's going on. It's returning when you retrieve by date range, but not when you pull up a single record. The view itself doesn't use the value there, as we have a `Participant` object loaded. I can add a null check.

stephenyeargin: Resolved with 9e09a1af

src/Pmi/Controller/BiobankController.php
```diff
@@ -253,6 +254,38 @@ public function ordersTodayAction(Application $app, Request $request)
...
+            $startString = '-' . intval($request->query->get('days')) . ' days';
+        }
+        $startTime = new \DateTime($startString, new \DateTimeZone($app->getUserTimezone()));
+        // Get MySQL date/time string in UTC
+        $startTime->setTimezone(new \DateTimezone('UTC'));
+        $today = $startTime->format('Y-m-d');
+
+        $quanumOrders = $app['pmi.drc.participants']->getOrders([
+            'startDate' => $today,
+            'endDate' => date('Y-m-d'),
```
jt2k, line 272: Summarizing conversation on Slack - because the RDR endpoint only filters on date and not date/time, we can't actually accomplish the goal of showing today's orders for the user's timezone. What we probably want to do is make sure that the end date is today in UTC (the latest possible date since we're only concerned about happenings in the western hemisphere) and use the start date of the user's time zone. This will get extra orders in some circumstances, but I think is better than missing some orders that should be considered "today".

stephenyeargin: Resolved with 07075bd

PR #649, opened 2020-04-23
[HPRO-608] Make the "Participant Details" tab for Cohort 3 available to HealthPro users

shyamnaru:


src/Pmi/Controller/DefaultController.php
```diff
@@ -394,7 +394,7 @@ public function participantAction($id, Application $app, Request $request)
         }
 
         $isCrossOrg = $participant->hpoId !== $app->getSiteOrganization();
-        $canViewDetails = !$isCrossOrg && ($participant->status || in_array($participant->statusReason, ['test-participant', 'basics']));
+        $canViewDetails = !$isCrossOrg && ($participant->status || in_array($participant->statusReason, ['test-participant', 'basics', 'genomics', 'ehr-consent']));
```
jt2k, line 397: The statusReason is set to the last check in the `parseRdrParticipant` method. Because of that, we should move `the `genomics` and `ehr-consent` checks above consent and withdrawal. That way, the blocking reasons are set last and take precedence.

PR #658, opened 2020-05-19
[HPRO-610] Add COPE surveys

jt2k:
HPRO-610

COPE surveys have been added to the participant details tab and the work queue.

Because the new survey columns need to be added at the end of the CSV export, we can no longer just loop through the surveys for the export, which resulted in this solution that stores the "initial" set of surveys in an array and then tacks on the new surveys manually at the end.

### DIFF
--- a/src/Pmi/Controller/WorkQueueController.php
+++ b/src/Pmi/Controller/WorkQueueController.php
@@ -318,8 +318,10 @@ class WorkQueueController extends AbstractController
                     'Completed Surveys'
                 ];
                 foreach (WorkQueue::$surveys as $survey => $label) {
-                    $headers[] = $label . ' PPI Survey Complete';
-                    $headers[] = $label . ' PPI Survey Completion Date';
+                    if (in_array($survey, WorkQueue::$initialSurveys)) {
+                        $headers[] = $label . ' PPI Survey Complete';
+                        $headers[] = $label . ' PPI Survey Completion Date';
+                    }
                 }
             } else {
                 $headers = [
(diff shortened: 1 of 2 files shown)


src/Pmi/Controller/WorkQueueController.php
```diff
@@ -447,6 +457,11 @@ public function exportAction(Application $app, Request $request)
                         $row[] = WorkQueue::dateFromString($participant->suspensionTime, $app->getUserTimezone());
                         $row[] = WorkQueue::csvStatusFromSubmitted($participant->consentForGenomicsROR);
                         $row[] = WorkQueue::dateFromString($participant->consentForGenomicsRORAuthored, $app->getUserTimezone());
+                        $row[] = WorkQueue::csvStatusFromSubmitted($participant->{"questionnaireOnCopeMay"});
+                        $row[] = WorkQueue::dateFromString($participant->{"questionnaireOnCopeMayAuthored"}, $app->getUserTimezone());
+                        $row[] = WorkQueue::csvStatusFromSubmitted($participant->{"questionnaireOnCopeJune"});
+                        $row[] = WorkQueue::dateFromString($participant->{"questionnaireOnCopeJuneAuthored"}, $app->getUserTimezone());$row[] = WorkQueue::csvStatusFromSubmitted($participant->{"questionnaireOnCopeJuly"});
```
shyamnaru, line 463: This can be in a new line `$row[] = WorkQueue::csvStatusFromSubmitted($participant->{"questionnaireOnCopeJuly"});`

PR #659, opened 2020-05-19
[HPRO-614] Work queue export fails when a patient status filter is set

shyamnaru:
This issue is we are calling session variable`$app->getSiteOrganizationId()` in `participantSummarySearch` function which is being looped while doing work queue export and causing the session start error.

### DIFF
--- a/src/Pmi/Controller/WorkQueueController.php
+++ b/src/Pmi/Controller/WorkQueueController.php
@@ -227,7 +227,9 @@ class WorkQueueController extends AbstractController
         //For ajax requests
         if ($request->isXmlHttpRequest()) {
             $params = array_merge($params, array_filter($request->request->all()));
-            $params['siteOrganizationId'] = $app->getSiteOrganizationId();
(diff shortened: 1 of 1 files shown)


src/Pmi/Controller/WorkQueueController.php
```diff
@@ -280,6 +281,7 @@ public function exportAction(Application $app, Request $request)
         $params = array_filter($request->query->all());
         $params['_count'] = $pageSize;
         $params['_sort:desc'] = 'consentForStudyEnrollmentAuthored';
+        $params['siteOrganizationId'] = $app->getSiteOrganizationId();
```
jt2k, line 284: Can we set this only if the patient status filter is set? Otherwise, we are adding this to every request, and the params array is logged with the export action.

jt2k: And for consistency, maybe make the same change for AJAX requests above, too.

PR #661, opened 2020-05-22
[HPRO-595] Allow approved Biobank users to finalize biobank orders in the Biobank view

shyamnaru:


src/Pmi/Service/NotifyBiobankOrderFinalizeService.php
```diff
@@ -0,0 +1,41 @@
...
+    protected $app;
+    protected $db;
+
+    public function __construct($app, $info)
+    {
+        $this->app = $app;
+        $this->info = $info;
+    }
+
+    public function sendEmails()
```
jt2k, line 19: Since the $this->info is specifically used by this method, would it make more sense to pass it in here instead of through the constructor? There's only one method right now, so it doesn't make a big difference, but I think it makes more sense for the constructor to only include the container.

views/biobank/order.html.twig
```diff
@@ -88,23 +88,42 @@
...
+
+            $('#form_finalized_ts').pmiDateTimePicker();
+
+            $('.finalize-form button[type="submit"]').on('click', function () {
+                //Display warning message
+                var message = 'Are you sure you want to finalize this order?';
+                $('input[name="form[finalized_samples][]"]').each(function () {
+                    //Select samples that are unchecked and not disabled
+                    if ($(this).prop('checked') === false && $(this).prop('disabled') === false) {
+                        message = 'Warning: At least one sample that was collected and processed (as applicable) was not finalized. Are you sure you wish to continue?';
```
jt2k, line 121: We might want to check to see if we want this warning to apply to the biobank users.

sql/healthpro/orders.sql
```diff
@@ -30,7 +30,9 @@ CREATE TABLE `orders` (
   `fedex_tracking` varchar(50) NULL DEFAULT NULL,
   `type` varchar(20) NULL DEFAULT NULL,
   `version` varchar(10) NULL DEFAULT NULL,
-  `history_id` int(11) DEFAULT NULL,
+  `history_id` int(11) DEFAULT NULL,,
+  `biobank` tinyint(1) NOT NULL DEFAULT 0,
```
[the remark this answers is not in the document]
shyamnaru: Sure!

PR #667, opened 2020-06-01
[HPRO-595] Fix send order to mayo issue

shyamnaru:


src/Pmi/Order/Order.php
```diff
@@ -1441,36 +1441,35 @@ public function loadFromJsonObject($object)
         return $this;
     }
 
-    public function sendOrderToMayo($participantId, $orderId, $app, $type = 'collected')
+    public function sendOrderToMayo($app, $info)
```
jt2k, line 1444: I might have missed something, but it looks like the app container should already be available from $this->app whenever this is called. Similarly, the participant is available because loadOrder has already been called (you removed it here, but in both cases where this method is called, loadOrder has to be called first).

As for the other fields passed in through info (collectedTs and finalizedSamples), is there a way to update the Order object first before calling this method so that this can do everything it needs to do without any additional parameters?

shyamnaru: Yes, you are right. Initially, this method used to be in OrderController and then I moved to Order class but didn't change the method params.

For other fields, I'll see if we can avoid passing those params. Thanks!

shyamnaru: Also, we are calling the `loadOrder` method to get the updated finalized_samples field from db. Since, I am passing those values thought we could remove calling that method. Moreover, we cannot call the` loadOrder` method in BiobankOrderController as we are not saving the finalized samples info into db at this point of time.

PR #669, opened 2020-06-03
[HPRO-595] Handle centrifuge type for kit orders

shyamnaru:


src/Pmi/Controller/BiobankController.php
```diff
@@ -194,7 +194,11 @@ public function orderAction($biobankId, $orderId, Application $app, Request $req
         $steps = ['collect', 'process', 'finalize'];
         // Set default step if current step not exists in the available steps
         $currentStep = !in_array($order->getCurrentStep(), $steps) ? 'collect' : $order->getCurrentStep();
-        $finalizeForm = $order->createBiobankOrderFinalizeForm($app['form.factory']);
+        $site = $app['em']->getRepository('sites')->fetchOneBy([
+            'deleted' => 0,
+            'google_group' => $app->getSiteId()
```
jt2k, line 199: Don't we want the order's created site here?

src/Pmi/Controller/BiobankController.php
```diff
@@ -225,13 +229,33 @@ public function orderAction($biobankId, $orderId, Application $app, Request $req
...
+                        }
+                        if ($order->get('type') === 'kit' && empty($order->get('processed_centrifuge_type'))) {
+                            $site = $app['em']->getRepository('sites')->fetchOneBy([
+                                'deleted' => 0,
+                                'google_group' => $app->getSiteId()
+                            ]);
+                            if (!empty($site['centrifuge_type'])) {
+                                $order->set('processed_centrifuge_type', $site['centrifuge_type']);
+                            }
+                        }
```
jt2k, line 248: 1. This if clause is the same as above. Not sure why it needs to be in a separate block.
2. Don't you already have the site retrieved from the database from above?

shyamnaru: Sorry, forgot to remove some of the unwanted code that I have written previously.

shyamnaru:
Also notice that in we should be getting the mayo account number using the current site id when `sendOrderToMayo` method is called in Order Controller (which I changed yesterday to fix the issue for biobank users).

PR #671, opened 2020-06-04
[HPRO-624] Add CSRF protection to order check form

shyamnaru:


src/Pmi/Controller/OrderController.php
```diff
@@ -61,6 +62,9 @@ public function orderCheckAction($participantId, Application $app)
 
     public function orderCreateAction($participantId, Application $app, Request $request)
     {
+        if ($request->request->has('order-check') && !$app['csrf.token_manager']->isTokenValid(new CsrfToken('orderCheck', $request->request->get('csrf_token')))) {
```
jt2k, line 65: The only way to get to this page is by POSTing the check form, right? Is there a reason why we shouldn't return a 403 whenever the CSRF check fails?

shyamnaru: No, we can also get here from order-create.html.twig which uses form builder `confirmForm` and results in empty csrf token.

jt2k: Ok. In that case, we want to the CSRF check any time the confirm form isn't being submitted. One way to do that is to put it after the confirm form is built. Then check for the confirm form not being submitted and the CSRF check failing. That way, all posts to this endpoint will be checking the CSRF token. The way you have it now, a POST directly to this endpoint without the `order-check` button or the confirmForm submit value would succeed.

PR #679, opened 2020-06-17
[HPRO-573] Allow scripps participant details tab for unpaired participants

shyamnaru:


src/Pmi/Controller/WorkQueueController.php
```diff
@@ -527,7 +527,7 @@ public function participantAction($id, Application $app, Request $request)
             $app->abort(404);
         }
 
-        if (!$app->hasRole('ROLE_AWARDEE_SCRIPPS') || !in_array($participant->awardee, $app->getAwardeeOrganization())) {
+        if (!$app->hasRole('ROLE_AWARDEE_SCRIPPS') || (!empty($participant->awardee) && !in_array($participant->awardee, $app->getAwardeeOrganization()))) {
```
jt2k, line 530: Shouldn't this also require a received 1SAL2 sample status? Also, let's try to make this clearer to avoid any confusion in the future.

Maybe separate the role check into it's own check so that it's clear:

```php
if (!$app->hasRole('ROLE_AWARDEE_SCRIPPS')) {
    $app->abort(403);
}
```

Then, add the more nuanced check for !\( \[awardee in allowed awardees\] OR \[unset awardee and 1SAL2 received\] \)

PR #681, opened 2020-06-18
[HPRO-629] Change "General Consent" to "Primary Consent"

shyamnaru:


stephenyeargin:
Any impact on the CSV exports?

shyamnaru:
Yes, as we are renaming it in CSV export as well. But, I think sites will be notified ahead regarding this change so it should be fine.

PR #685, opened 2020-06-26
[HPRO-599] Symfony adjustments

jt2k:
* Disables routing to Symfony in stable and prod
* Removes the need to have two copies of the DatastoreSessionHandler (the other services have modifications that prevents us from reusing)


web/index.php
```diff
@@ -1,8 +1,13 @@
 <?php
+require_once __DIR__ . '/../vendor/autoload.php';
 
-$url = @parse_url($_SERVER['REQUEST_URI'])['path'];
+use Pmi\Application\HpoApplication;
 
-if (preg_match("/^\/s(\/|$)/", $url)) {
+$app = new HpoApplication();
+$allowSymfony = !($app->isStable() || $app->isProd());
```
stephenyeargin, line 7: This would be temporary, right? Thinking at a comment block above it might be useful.

jt2k: Comments added.

PR #686, opened 2020-06-29
[HPRO-635] Handle EARLY_OUT withdrawal status

stephenyeargin:
This PR handles the work queue and participant views if for some reason the EARLY_OUT status were to slip through the HealthPro UI.

[HPRO-635]

[HPRO-635]: https://precisionmedicineinitiative.atlassian.net/browse/HPRO-635

### DIFF
--- a/src/Pmi/Entities/Participant.php
+++ b/src/Pmi/Entities/Participant.php
@@ -20,6 +20,7 @@ class Participant
     public $isCoreParticipant = false;
     public $activityStatus;
     public $isSuspended = false;
+    public $isWithdrawn = false;
     public $consentCohortText;
 
     private $disableTestAccess;
@@ -103,6 +104,7 @@ class Participant
         if (!empty($participant->withdrawalStatus) && in_array($participant->withdrawalStatus, ['NO_USE', 'EARLY_OUT'], true)) {
             $this->status = false;
             $this->statusReason = 'withdrawal';
+            $this->isWithdrawn = true;
         }
 
         // Map gender identity to gender options for MayoLINK.
--- a/src/Pmi/Entities/Participant.php
+++ b/src/Pmi/Entities/Participant.php
@@ -34,6 +34,11 @@ class Participant
         'COHORT_3' => 'Cohort 3'
     ];
 
+    private static $withdrawalStatusValues = [
+        'NO_USE',
+        'EARLY_OUT'
+    ];
+
     public function __construct($rdrParticipant = null)
     {
         if (is_object($rdrParticipant)) {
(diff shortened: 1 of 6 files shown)


jt2k:
@stephenyeargin Any thoughts on my comment in the ticket description?

> [One way] to approach this is to calculate an `isWithdrawn` boolean status, similar to how we handle `isSuspended`. Then, anywhere we check the participant's `withdrawalStatus` for `NO_USE`, we can switch to using the `isWithdrawn` value.

src/Pmi/Entities/Participant.php
```diff
@@ -338,7 +340,7 @@ private function getSiteSuffix($site)
 
     private function getActivityStatus($participant)
     {
-        if ($participant->withdrawalStatus === 'NO_USE') {
+        if (in_array($participant->withdrawalStatus, ['NO_USE', 'EARLY_OUT'], true)) {
```
stephenyeargin, line 343: I wasn't able to centralize the logic for this one because `getActivityStatus()` is taking a `StdClass` value rather than a `Participant`.

src/Pmi/WorkQueue/WorkQueue.php
```diff
@@ -340,7 +340,8 @@ public function generateTableRows($participants, $app)
             $row['dvEhrStatus'] = $this->displayConsentStatus($participant->consentForDvElectronicHealthRecordsSharing, $participant->consentForDvElectronicHealthRecordsSharingAuthored);
             $row['caborConsent'] = $this->displayConsentStatus($participant->consentForCABoR, $participant->consentForCABoRAuthored);
             $row['activityStatus'] = $this->getActivityStatus($participant);
-            $row['withdrawalStatus'] = $participant->withdrawalStatus; // Used to add withdrawn class in the data tables
+            $row['withdrawalStatus'] = $participant->withdrawalStatus;
```
stephenyeargin, line 343: This can _likely_ be dropped from the payload if the prior comment is correct that it is only used for formatting.

web/local-router.php
```diff
@@ -5,7 +5,7 @@
 } elseif (preg_match('/\\.pdf|csv|json$/', $_SERVER["REQUEST_URI"])) {
     // for non-asset PDF, CSV and JSON URLs, explicitly send to index.php controller to avoid
     // the dev server from assuming these are static assets
-    require 'silex-index.php';
+    require 'index.php';
```
stephenyeargin, line 8: Cherry-picked change from the `develop` branch.

stephenyeargin:
> @stephenyeargin Any thoughts on my comment in the ticket description?

Adjusted. I'd jumped into the code without reading all the way down. 🐼

src/Pmi/WorkQueue/WorkQueue.php
jt2k: I agree that this can be removed. The only place it was being used was in the `createdRow` callback in workqueue.js which you've already modified to use the new field.

src/Pmi/Entities/Participant.php
jt2k: Makes sense. Maybe put these two values in a static class array so that it can be used in both places?

stephenyeargin: Done in 30ca5ba

PR #689, opened 2020-07-08
[HPRO-633] Allow users to bulk upload patient status information in HealthPro

shyamnaru:


src/Pmi/Service/PatientStatusService.php
```diff
@@ -28,29 +29,75 @@ public function sendPatientStatusToRdr()
...
+                // RDR status
+                // 2 = RDR 400 (Invalid participant id)
+                // 3 = RDR 500 (Invalid patient status)
+                // 4 = Other RDR errors
+                $rdrStatus = 4;
+                if ($this->rdr->getLastErrorCode() === 400) {
+                    $rdrStatus = 2;
+                } elseif ($this->rdr->getLastErrorCode() === 500) {
+                    $rdrStatus = 3;
+                }
```
jt2k, line 69: A few comments on the status:

1. Can use class constants (maybe in the `PatientStatusHistory` entity class) to reference these integers? That way we don't have to use these arbitrary status values throughout the code and it is more immediately obvious what the statuses mean.
2. We don't know for sure that a 500 error is always going to be because of an invalid patient status. So the comment here (or class constant if (1) is implemented) should reflect that.

symfony/src/Controller/PatientStatusController.php
```diff
@@ -0,0 +1,176 @@
...
+            if ($form->isValid()) {
+                if (!empty($patientStatuses)) {
+                    $organization = $em->getRepository(Organizations::class)->findOneBy(['id' => $session->get('siteOrganizationId')]);
+                    $patientStatusImport = new PatientStatusImport();
+                    $patientStatusImport->setFileName($fileName);
+                    $patientStatusImport->setOrganization($organization);
+                    $patientStatusImport->setAwardee($session->get('siteAwardeeId'));
+                    $patientStatusImport->setUserId($this->getUser()->getId());
+                    $patientStatusImport->setSite($session->get('site')->id);
+                    $patientStatusImport->setCreatedTs(new \DateTime());
```
jt2k, line 49: Minor style suggestion (would apply to other places as well)

```suggestion
                    $patientStatusImport = new PatientStatusImport();
                    $patientStatusImport
                        ->setFileName($fileName)
                        ->setOrganization($organization)
                        ->setAwardee($session->get('siteAwardeeId'))
                        ->setUserId($this->getUser()->getId())
                        ->setSite($session->get('site')->id)
                        ->setCreatedTs(new \DateTime());
```

symfony/src/Controller/PatientStatusController.php
```diff
@@ -0,0 +1,176 @@
...
+            throw $this->createNotFoundException('Page Not Found!');
+        }
+        //For ajax requests
+        if ($request->isXmlHttpRequest()) {
+            $params = $request->request->all();
+            $patientStatusHistories = $patientStatusImport->getPatientStatusHistories()->slice($params['start'], $params['length']);
+            $ajaxData = [];
+            $ajaxData['data'] = $patientStatusImportService->getAjaxData($patientStatusHistories, $patientStatusImport->getOrganization());
+            $ajaxData['recordsTotal'] = $ajaxData['recordsFiltered'] = count($patientStatusImport->getPatientStatusHistories());
+            return new JsonResponse($ajaxData, 200);
```
jt2k, line 171: Default response code is 200. Not sure of the value of explicitly setting it here.

symfony/src/Entity/PatientStatus.php
```diff
@@ -0,0 +1,134 @@
...
+     * @ORM\Id()
+     * @ORM\GeneratedValue()
+     * @ORM\Column(type="integer")
+     */
+    private $id;
+
+    /**
+     * @ORM\Column(type="string", length=50)
+     */
+    private $participant_id;
```
jt2k, line 24: Applies here and in all of the entities. Property names should be camel-cased. Doctrine will automatically use the corresponding underscore-separated column name.

I just looked it up and realized that PSR-2 purposely doesn't include this as a SHOULD/MUST, but it is the default for how auto-generated entities are created and is the standard that we currently use in our other classes.

symfony/src/Controller/PatientStatusController.php
```diff
@@ -0,0 +1,180 @@
...
+                        ->setPatientStatus($patientStatus)
+                        ->setImport($patientStatusImport);
+                    $em->persist($patientStatusHistory);
+
+                    // Update history id in patient_status table
+                    $patientStatus->setHistory($patientStatusHistory);
+                    if (($key % $batchSize) === 0) {
+                        $em->flush();
+                        $loggerService->log(Log::PATIENT_STATUS_HISTORY_ADD, $patientStatusHistory->getId());
+                        $loggerService->log(Log::PATIENT_STATUS_EDIT, $patientStatus->getId());
```
jt2k, line 127: Do we need these log entries? We are already logging the patient status import entry, and each history item is connected to the import entry. Also, the way it is written here, it will only log the last status and status history action in each batch.

Another factor is that we haven't implemented the buffered logger yet, so this would generate a bunch of remote calls to Stackdriver.

src/Pmi/Service/PatientStatusService.php
```diff
@@ -22,35 +25,75 @@ public function sendPatientStatusToRdr()
...
-            LEFT JOIN users u ON psh.user_id = u.id
-            WHERE psh.rdr_ts is null
-            ORDER BY psh.id ASC
+            FROM patient_status_import_rows psir
+            INNER JOIN patient_status_import psi ON psi.id = psir.import_id AND psi.confirm = :confirm
+            LEFT JOIN patient_status ps ON ps.participant_id = psir.participant_id AND psi.organization = ps.organization
+            LEFT JOIN users u ON psi.user_id = u.id
+            WHERE psir.rdr_status = :rdrStatus
+            ORDER BY psir.id ASC
             limit {$limit}
```
jt2k, line 42: I know this was already here before, but can we use a parameter for the limit? While we have control over the config items, we should treat them like untrusted input.

symfony/src/Controller/DefaultController.php
```diff
@@ -16,7 +17,7 @@ class DefaultController extends AbstractController
      */
     public function index(LoggerService $loggerService)
     {
-        $loggerService->log('REQUEST', 'HealthPro Symfony Home Page');
+        $loggerService->log(Log::REQUEST, 'HealthPro Symfony Home Page');
```
jt2k, line 20: Let's remove this, now that we've straightened out the logging

src/Pmi/Service/PatientStatusService.php
```diff
@@ -22,35 +25,75 @@ public function sendPatientStatusToRdr()
...
+                $this->app->log(Log::PATIENT_STATUS_IMPORT_EDIT, $importId);
             }
         }
     }
+
+    public function deleteUnconfirmedImportData()
+    {
+        $date = (new \DateTime('UTC'))->modify('-1 hours');
+        $date = $date->format('Y-m-d H:i:s');
+        $this->db->query("DELETE psir FROM patient_status_import_rows psir inner join patient_status_import psi on psir.import_id = psi.id where psi.created_ts < '{$date}' and psi.confirm = 0");
```
jt2k, line 97: Should we delete the patient_status_import row, too?

src/Pmi/Service/PatientStatusService.php
shyamnaru: Somehow setting parameter to a limit clause throws a syntax error. It may be related to this https://github.com/doctrine/dbal/issues/3459

I tried setting it to `$patientStatuses = $this->em->fetchAll($query, ['confirm' => 1, 'rdrStatus' => 0, 'limit' => $limit], [\PDO::PARAM_INT, \PDO::PARAM_INT, \PDO::PARAM_INT]);` but that didn't work either.

src/Pmi/Service/PatientStatusService.php
shyamnaru: I thought we could track the exports that are not confirmed but I am fine with deleting.

src/Pmi/Service/PatientStatusService.php
jt2k: Oh, right. We should just cast the setting to an integer. And maybe abort if it is < 1. No parameter needed in that case.

src/Pmi/Service/PatientStatusService.php
jt2k: If that's intentional, that's fine with me. I just wanted to make sure it would never show up for the user after the import rows have been deleted.

symfony/src/Service/PatientStatusImportService.php
```diff
@@ -0,0 +1,78 @@
...
+        }
+        // Check column headers
+        foreach ($headers as $header) {
+            // Handle bom
+            $header = str_replace("\xEF\xBB\xBF", '', $header);
+            if (!in_array(str_replace(' ', '', strtolower($header)), $validHeaders)) {
+                $form['patient_status_csv']->addError(new FormError("Invalid column headers"));
+                return;
+            }
+        }
```
jt2k, line 37: I'm not sure we need to enforce specific strings in the header row. In the instructions, these three valid headers are never mentioned. I think we should just check that the first row returns something and continue.

jt2k:
@shyamnaru Hope you don't mind. I had some minor layout changes to suggest for the import page and form that I decided was easier to just commit directly.

symfony/src/Entity/PatientStatus.php
```diff
@@ -0,0 +1,134 @@
...
+     */
+    private $organization;
+
+    /**
+     * @ORM\Column(type="string", length=50)
+     */
+    private $awardee;
+
+    /**
+     * @ORM\OneToMany(targetEntity="App\Entity\PatientStatusHistory", mappedBy="patient_status")
```
jt2k, line 37: This is throwing an invalid entity mapping error. The mapped by should refer to the entity name, not the column name.

```suggestion
     * @ORM\OneToMany(targetEntity="App\Entity\PatientStatusHistory", mappedBy="patientStatus")
```

symfony/src/Service/PatientStatusImportService.php
shyamnaru: Sure!

symfony/src/Entity/PatientStatus.php
shyamnaru: This might have been missed while changing entity names to camel case. Will fix that. Thanks!

PR #692, opened 2020-07-14
[HPRO-599] Adjust Symfony logging

jt2k:
HPRO-599

This PR adjusts the logging in Symfony a bit. It still doesn't match our Silex implementation exactly, but it is closer.

I'm taking the shortcut of just using Symfony's dev vs. prod environment and not worrying about the differences between our environments. For now, we can assume that when running locally we will be using Symfony's dev environment and when running in App Engine, it will be using the prod environment.

To that end, I'm removing the default prod.log file and relying on syslog and Stackdriver only in production. In dev, I have removed the Stackdriver handler, as we don't want this on by default. In Silex we have a local configuration option that can be used to test. To test locally in the Symfony, just add the handler temporarily back to the dev monolog configuration.

This PR also adds a request listener to add our custom log entry on every request, which adds an info-level Stackdriver log with the user information and context.

### DIFF
--- a/symfony/src/EventListener/RequestListener.php
+++ b/symfony/src/EventListener/RequestListener.php
@@ -1,0 +1,30 @@
+<?php
+
+namespace App\EventListener;
+
+use Symfony\Component\HttpKernel\Event\RequestEvent;
+use App\Service\LoggerService;
+
+class RequestListener
+{
+    private $logger;
+
+    public function __construct(LoggerService $logger)
+    {
+        $this->logger = $logger;
+    }
+
+    public function onKernelRequest(RequestEvent $event)
+    {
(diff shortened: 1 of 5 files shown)


symfony/src/EventListener/RequestListener.php
```diff
@@ -0,0 +1,30 @@
...
+    private $logger;
+
+    public function __construct(LoggerService $logger)
+    {
+        $this->logger = $logger;
+    }
+
+    public function onKernelRequest(RequestEvent $event)
+    {
+        if (!$event->isMasterRequest()) {
```
stephenyeargin, line 19: Even after reading [the docs](https://symfony.com/doc/current/event_dispatcher.html) on this, I'm still not sure about sub requests.

PR #693, opened 2020-07-14
[HPRO-599] Add Symfony support for page notices

jt2k:
HPRO-599

Implements the page notice system in the Symfony part of our application.

### DIFF
--- a/symfony/src/Entity/Notice.php
+++ b/symfony/src/Entity/Notice.php
@@ -11,9 +11,9 @@ use Doctrine\ORM\Mapping as ORM;
 class Notice
 {
     /**
-     * @ORM\Column(name="id", type="integer", nullable=false)
      * @ORM\Id
-     * @ORM\GeneratedValue(strategy="IDENTITY")
+     * @ORM\Column(name="id", type="integer", nullable=false)
+     * @ORM\GeneratedValue()
      */
     private $id;
 
--- a/symfony/src/Entity/Notice.php
+++ b/symfony/src/Entity/Notice.php
@@ -1,0 +1,126 @@
+<?php
+
+namespace App\Entity;
+
+use Doctrine\ORM\Mapping as ORM;
+
+/**
+ * @ORM\Entity(repositoryClass="App\Repository\NoticeRepository")
+ * @ORM\Table(name="notices")
+ */
+class Notice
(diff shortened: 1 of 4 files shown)


symfony/src/Entity/Notice.php
```diff
@@ -0,0 +1,126 @@
...
+/**
+ * @ORM\Entity(repositoryClass="App\Repository\NoticeRepository")
+ * @ORM\Table(name="notices")
+ */
+class Notice
+{
+    /**
+     * @ORM\Column(name="id", type="integer", nullable=false)
+     * @ORM\Id
+     * @ORM\GeneratedValue(strategy="IDENTITY")
```
stephenyeargin, line 16: Not sure if this is generated code, but was curious whether we had to specify column names and the `GeneratedValue` strategy for our stuff?

symfony/src/EventListener/RequestListener.php
```diff
@@ -20,11 +30,36 @@ public function onKernelRequest(RequestEvent $event)
...
     }
+
+    private function checkPageNotices()
+    {
+        $path = $this->request->getPathInfo();
+        $activeNotices = $this->em->getRepository(Notice::class)
+            ->getActiveNotices($path);
+
+        // If one of the notices is a full page notice, render the notice response and return
+        if (!preg_match('/^(\/s)?\/admin/', $path)) { // Ignore full page notices for admin urls
```
stephenyeargin, line 53: If we ever drop the `/s` post-migration, we'll want to adjust this.

jt2k: The `/s` is optional, so it will match both `/admin...` and `/s/admin...`. But yes, after we completely migrate, we can drop the optional `/s`.

symfony/src/Entity/Notice.php
jt2k: I auto-generated this entity using the `doctrine:mapping:import` command. I'll look into whether we should remove the explicit strategy parameter.

symfony/src/EventListener/RequestListener.php
stephenyeargin: :+1: Some day I'll learn how to reed a regex. 😖

PR #698, opened 2020-07-20
[HPRO-644] Suppress router listener logs in Symfony

jt2k:
HPRO-644

Accomplished this by overriding the router_listener service definition and passing in `null` for the logger.

Also added a few helper scripts into the `bin` directory.

### DIFF
--- a/bin/dx
+++ b/bin/dx
@@ -1,0 +1,10 @@
+#!/bin/bash
+#
+# Wrapper for docker exec to run a command inside of the web container
+#
+# Example:
+# bin/dx bin/phpunit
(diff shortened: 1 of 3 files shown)


bin/dx
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+#
+# Wrapper for docker exec to run a command inside of the web container
+#
+# Example:
+# bin/dx bin/phpunit
+# bin/dx bin/symfony cache:clear
+#
+
+docker exec -it healthpro-web "$@"
```
stephenyeargin, line 10: 😻

PR #703, opened 2020-07-24
[HPRO-650] Migrate 'Page Notices' to Symfony

stephenyeargin:
This takes a relatively contained piece of functionality of the site (the Page Notices tool) and migrates it into the Symfony framework. Among the things to discuss:

- Maintaining routes between the two frameworks (use static mapping file?)
- Checking for permissions/security (controller vs. firewall route matching)
- Migrating form types backed by Entities
- Maintaining the two `base.html.twig` files
- Route naming conventions
- Twig globals (e.g. `app.userTimezoneDisplay`)

[HPRO-650]

[HPRO-650]: https://precisionmedicineinitiative.atlassian.net/browse/HPRO-650

### DIFF
--- a/src/Pmi/Controller/AdminController.php
+++ b/src/Pmi/Controller/AdminController.php
@@ -40,11 +40,6 @@ class AdminController extends AbstractController
         ['deactivateNotifications', '/notifications/deactivate'],
         ['missingMeasurements', '/missing/measurements', ['method' => 'GET|POST']],
         ['missingOrders', '/missing/orders', ['method' => 'GET|POST']],
-        ['notices', '/notices'],
-        ['notice', '/notice/{id}', [
-            'method' => 'GET|POST',
-            'defaults' => ['id' => null]
-        ]],
         ['patientStatusRdrJson', '/patientstatus/{participantId}/organization/{organizationId}/rdr.json', ['method' => 'GET']],
         ['patientStatusHistoryRdrJson', '/patientstatus/{participantId}/organization/{organizationId}/history/rdr.json', ['method' => 'GET']],
         ['participants', '/testing/participants', ['method' => 'GET|POST']],
@@ -421,66 +416,6 @@ class AdminController extends AbstractController
         }
         return $app['twig']->render('admin/missing/orders.html.twig', [
             'missing' => $missing,
-            'form' => $form->createView()
-        ]);
-    }
-
-    public function noticesAction(Application $app)
-    {
-        $notices = $app['em']->getRepository('notices')->fetchBy([], ['id' => 'asc']);
-        return $app['twig']->render('admin/notices/index.html.twig', [
-            'notices' => $notices
-        ]);
-    }
-
-    public function noticeAction($id, Application $app, Request $request)
-    {
-        if ($id) {
-            $notice = $app['em']->getRepository('notices')->fetchOneBy(['id' => $id]);
-            if (!$notice) {
-                $app->abort(404);;
-            }
-        } else {
-            $notice = null;
-        }
-
-        $form = $app['form.factory']->createNamed(
-            'form',
-            NoticeType::class,
-            $notice ?: ['status' => 1],
-            ['timezone' => $app->getUserTimezone()]
-        );
-
-        $form->handleRequest($request);
-        if ($form->isSubmitted()) {
-            if ($form->isValid()) {
-                if ($notice === null) {
-                    if ($id = $app['em']->getRepository('notices')->insert($form->getData())) {
-                        $app->log(Log::NOTICE_ADD, $id);
-                        $app->addFlashNotice('Notice added');
-                    }
-                } elseif ($request->request->has('delete')) {
-                    if ($app['em']->getRepository('notices')->delete($id)) {
-                        $app->log(Log::NOTICE_DELETE, $id);
-                        $app->addFlashNotice('Notice removed');
-                    }
-                } else {
-                    if ($app['em']->getRepository('notices')->update($id, $form->getData())) {
-                        $app->log(Log::NOTICE_EDIT, $id);
-                        $app->addFlashNotice('Notice updated');
-                    }
-                }
-                return $app->redirectToRoute('admin_notices');
-            } else {
-                // Add a form-level error if there are none
-                if (count($form->getErrors()) == 0) {
-                    $form->addError(new FormError('Please correct the errors below'));
-                }
-            }
-        }
-
-        return $app['twig']->render('admin/notices/edit.html.twig', [
-            'notice' => $notice,
             'form' => $form->createView()
         ]);
     }
--- a/symfony/config/routes.yaml
+++ b/symfony/config/routes.yaml
@@ -1,3 +1,50 @@
 index:
     path: /
     controller: App\Controller\DefaultController::index
+
+# Legacy Silex Routes
+
+settings:
+    path: /settings
+
+logout:
+    path: /logout
+
+help_home:
+    path: /help
+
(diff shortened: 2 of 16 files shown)


symfony/src/Controller/NoticeController.php
```diff
@@ -0,0 +1,84 @@
...
+
+    /**
+     * @Route("/notice/{id}", name="admin_notice")
+     */
+    public function edit(NoticeRepository $noticeRepository, EntityManagerInterface $em, LoggerService $loggerService, Request $request, $id=null)
+    {
+        if ($id) {
+            $notice = $noticeRepository->find($id);
+            if (!$notice) {
+                $this->createNotFoundException('Page notice not found.');
```
stephenyeargin, line 39: This needs a `throw`.

symfony/templates/notice/edit.html.twig
```diff
@@ -32,7 +32,7 @@
                 </div>
                 <div class="panel-body">
                     {% set return = path('admin_notice', { notice: notice ? notice.id : null }) %}
-                    <p><strong>Your time zone</strong>: {{ app.userTimezoneDisplay }} <a href="{{ path('settings', { return: return}) }}" class="small">change</a></p>
+                    <p><strong>Your time zone</strong>: {{ app.user.info.timezone|default('Not set!') }} <a href="{{ path('settings', { return: settings_return_url }) }}" class="small">change</a></p>
```
stephenyeargin, line 35: This displays `America/Chicago` instead of `Central Time`. To maintain one-for-one functionality, the lookup method either needs to be added/referenced to the `User` entity or elsewhere.

views/admin/index.html.twig
```diff
@@ -17,7 +17,7 @@
             <a href="{{ path('admin_withdrawalNotifications') }}" class="btn btn-lg btn-default btn-block"><i class="fa fa-bell" aria-hidden="true"></i> Withdrawal Notifications</a>
         </div>
         <div class="col-sm-6 col-md-4">
-            <a href="{{ path('admin_notices') }}" class="btn btn-lg btn-default btn-block"><i class="fa fa-info-circle" aria-hidden="true"></i> Page Notices</a>
+            <a href="/s/admin/notices" class="btn btn-lg btn-default btn-block"><i class="fa fa-info-circle" aria-hidden="true"></i> Page Notices</a>
```
stephenyeargin, line 20: This is a case of Silex needing to know the Symfony route.

symfony/templates/notice/index.html.twig
```diff
@@ -19,9 +19,9 @@
             <tr>
                 <td><a href="{{ path('admin_notice', { id: notice.id }) }}">{{ notice.url }}</a></td>
                 <td>{{ notice.message|length > 100 ? notice.message|slice(0, 100) ~ '...' : notice.message  }}</td>
-                <td>{{ notice.full_page ? 'Yes' : 'No' }}</td>
-                <td>{{ notice.start_ts ? notice.start_ts|date('n/j/Y g:ia', app.userTimezone) }}</td>
-                <td>{{ notice.end_ts ? notice.end_ts|date('n/j/Y g:ia', app.userTimezone) }}</td>
+                <td>{{ notice.fullPage ? 'Yes' : 'No' }}</td>
+                <td>{{ notice.startTs ? notice.startTs|date('n/j/Y g:ia', app.userTimezone) }}</td>
+                <td>{{ notice.endTs ? notice.endTs|date('n/j/Y g:ia', app.userTimezone) }}</td>
```
jt2k, line 24: Don't we need to switch these to `app.user.info.timezone`?

symfony/src/Controller/NoticeController.php
```diff
@@ -0,0 +1,84 @@
...
+use Doctrine\ORM\EntityManagerInterface;
+use Pmi\Audit\Log;
+use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
+use Symfony\Component\Form\FormError;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\Routing\Annotation\Route;
+
+/**
+ * @Route("/s/admin/notices")
+ */
```
jt2k, line 17: Need to require ROLE_ADMIN for this controller. Or add all of `/admin` to `access_control` in `security.yaml`.

src/Pmi/Controller/AdminController.php
```diff
@@ -425,66 +420,6 @@ public function missingOrdersAction(Application $app, Request $request, $_route)
...
-                $app->abort(404);;
-            }
-        } else {
-            $notice = null;
-        }
-
-        $form = $app['form.factory']->createNamed(
-            'form',
-            NoticeType::class,
-            $notice ?: ['status' => 1],
```
jt2k, line 450: This functionality didn't make it to the new controller. When creating a new notice, the enabled status should default to true. (i.e. create a new Notice instance and set the status instead of passing null to the form here: https://github.com/all-of-us/healthpro/pull/703/files#diff-0131fcea65cb24fc2f156786d74ade40R42)

symfony/src/Controller/NoticeController.php
```diff
@@ -0,0 +1,84 @@
...
+        if ($id) {
+            $notice = $noticeRepository->find($id);
+            if (!$notice) {
+                throw $this->createNotFoundException('Page notice not found.');
+            }
+        } else {
+            $notice = null;
+        }
+
+        $form = $this->createForm(NoticeType::class, $notice, ['timezone' => $this->getUser()->getTimezone()]);
```
jt2k, line 45: This isn't working because the underlying `getTimezone` method is currently broken (and unused). We need to either get this from the info array or fix the User class to get the timezone directly.

symfony/config/routes.yaml
```diff
@@ -1,3 +1,50 @@
 index:
     path: /
     controller: App\Controller\DefaultController::index
+
+# Legacy Silex Routes
```
jt2k, line 5: Nice!

symfony/src/Controller/NoticeController.php
stephenyeargin: Do we have opinions on using `FrameworkExtraBundle` so it can be an annotation, or would we rather explicitly call it in PHP?

jt2k: My opinion is to either use annotations or do pattern matching in the `security.access_control` configuration. Since Silex is currently doing pattern matching, my initial thought is to do that for now, but I'm open to other preferences.

stephenyeargin: I'd agree. We may want the annotations for later work where the permission structure is more complex (and to keep the code centralized), but in the name of keeping this PR simple, I'll update the YAML file.

symfony/src/Controller/NoticeController.php
stephenyeargin: Should be resolved as part of 39524cb.

src/Pmi/Controller/AdminController.php
stephenyeargin: Resolved in d8bb5a2.

symfony/src/Controller/NoticeController.php
stephenyeargin: Fixed in 3059fff. Also migrated the error page(s) over. Symfony lets you split them out by code, but I used the fallback template for now.

symfony/templates/notice/index.html.twig
stephenyeargin: Switched them to `app.user.timezone` with the fix in 39524cb.

PR #710, opened 2020-08-06
[HPRO-656] Fix physical measurements FHIR bundle error

shyamnaru:


src/Pmi/Evaluation/Evaluation.php
```diff
@@ -695,4 +676,28 @@ public function getReasonDisplayText()
         $reasonDisplayText = array_search($this->evaluation['eh_reason'], self::$cancelReasons);
         return !empty($reasonDisplayText) ? $reasonDisplayText : 'Other';
     }
+
+    private function addDiastolicBloodPressureConstraint($form, $field)
```
jt2k, line 680: While the only field where we currently use the `compare` option is diastolic blood pressure, there's nothing in this functionality that is specific to this measure. Can we rename this to something like `addCompareConstraint`?

PR #720, opened 2020-08-25
[HPRO-658] New "Today's Participants" Features

shyamnaru:


src/Pmi/Controller/ReviewController.php
```diff
@@ -30,19 +33,52 @@ public function todayAction(Application $app, Request $request)
...
+
         // Get beginning of today (at midnight) in user's timezone
-        $startString = 'today';
-        // Allow overriding start time to test in non-prod environments
-        if (!$app->isProd() && intval($request->query->get('days')) > 0) {
-            $startString = '-' . intval($request->query->get('days')) . ' days';
+        $startDate = new \DateTime('today', new \DateTimeZone($app->getUserTimezone()));
+
+        // Get end of today (at midnight) in user's timezone
+        $endDate = new \DateTime('yesterday 1 sec ago', new \DateTimeZone($app->getUserTimezone()));
```
jt2k, line 42: I think we can simplify a lot of the date math by:

1. Setting the form's view timezone to the local timezone
2. Changing the query to use `< :endTime` instead of `<=` which allows us to just use midnight of the following day instead of `23:59:59` of the specified day

See diff from this branch:
https://github.com/all-of-us/healthpro/compare/sn/hpro-658/new-todays-participants...jt/HPRO-658/todays-participants-simplify-date-math

jt2k: I also made some minor text changes to the error message and display message.

views/review/today.html.twig
```diff
@@ -30,132 +40,138 @@
         <li role="presentation"><a href="{{ path('review_measurementsRecentModify') }}">Recently Modified Physical Measurements</a></li>
     </ul>
```
jt2k, line 41: Can we make the Today's Participant tab clickable (i.e. add the link to the href) so that the user can click on the tab to reset to today's view?

PR #724, opened 2020-09-02
[HPRO-630] Report deceased participants

stephenyeargin:
Allows users to report Deceased Participants as well as review those submissions for accuracy.

[HPRO-630]

[HPRO-630]: https://precisionmedicineinitiative.atlassian.net/browse/HPRO-630


symfony/src/Entity/DeceasedReport.php
```diff
@@ -0,0 +1,307 @@
+<?php
+
+namespace App\Entity;
+
+use Symfony\Component\Validator\Constraints as Assert;
+
+class DeceasedReport
```
stephenyeargin, line 7: Even though this is an API resource, it made sense to me to encapsulate it in a class for easier portability (and less code repetition).

symfony/src/Service/EnvironmentService.php
```diff
@@ -126,4 +127,47 @@ public function getUserTimezone($useDefault = true)
             return null;
         }
     }
+
+    protected function loadConfiguration($override = [])
```
stephenyeargin, line 131: This will need testing with the emulator. I haven't worked a lot with it in Docker.

symfony/src/Service/RdrApiService.php
```diff
@@ -0,0 +1,72 @@
...
+namespace App\Service;
+
+use App\Service;
+use Pmi\Entities\Configuration as DatastoreConfiguration;
+use Google_Client as GoogleClient;
+use Google_Service_Oauth2 as GoogleServiceOauth2;
+use Pmi\HttpClient;
+use Symfony\Component\HttpKernel\KernelInterface;
+
+class RdrApiService
```
stephenyeargin, line 12: In most cases, this would be injected into a separate class (e.g. `ParticipantSummaryService` or `DeceasedReportsService`) rather than called directly from a controller.

symfony/src/Twig/AppExtension.php
```diff
@@ -22,7 +29,10 @@ public function getFunctions()
             new TwigFunction('path_exists', [$this, 'checkPath']),
             new TwigFunction('asset', [$this, 'asset']),
             new TwigFunction('slugify', [$this, 'slugify']),
-            new TwigFunction('timezone_display', [$this, 'timezoneDisplay'])
+            new TwigFunction('timezone_display', [$this, 'timezoneDisplay']),
+            new TwigFunction('codebook_display', [$this, 'getCodeBookDisplay']),
+            new TwigFunction('organization_display', [$this, 'getAwardeeDisplay']),
+            new TwigFunction('awardee_display', [$this, 'getAwardeeDisplay'])
```
stephenyeargin, line 35: The intent is to replicate what we were doing in `HpoApplication`. Both `getAwardeeDisplay` and `getOrganizationDisplay` likely need some level of caching in case the lookup happens in a loop.

stephenyeargin: Opted to cache it in memory.

symfony/src/Entity/DeceasedReport.php
```diff
@@ -0,0 +1,315 @@
...
+                    break;
+                case 'https://www.pmi-ops.org/observation/reviewed':
+                    $this->setReviewedBy($performer->reference);
+                    $this->setReviewedOn(new \DateTime($performer->extension[0]->valueDateTime));
+                    break;
+            }
+
+        }
+
+        if (property_exists($report->extension[0], 'valueHumanName') && $report->extension[0]->valueHumanName) {
```
shyamnaru, line 307: Got a 500 error as the `$report-extension` is undefined. I think here we also need to check if the `$report->extension` is defined.

symfony/src/Entity/DeceasedReport.php
```diff
@@ -0,0 +1,315 @@
...
+            $this->setNextOfKinName($report->extension[0]->valueHumanName->text);
+            $this->setNextOfKinRelationship($report->extension[0]->valueHumanName->extension[0]->valueCode);
+            $this->setNextOfKinTelephoneNumber($report->extension[0]->valueHumanName->extension[1]->valueString);
```
shyamnaru, line 310: Similarly, I got an error as the index 1 is not defined `valueHumanName->extension[1]`. I think we need make sure that these indexes are defined.

symfony/src/Entity/DeceasedReport.php
stephenyeargin: Going to rework this so that it uses the same 'check the type rather than the index' method I did for the submitter/reviewer above.

symfony/src/Entity/DeceasedReport.php
stephenyeargin: Fixed in 4409e11

web/assets/js/app.js
```diff
@@ -234,9 +234,9 @@ $(document).ready(function()
     /*************************************************************************
      * Plugin to initialize datetimepicker and register change event listener
      ************************************************************************/
-    $.fn.pmiDateTimePicker = function(dateFormat = null) {
+    $.fn.pmiDateTimePicker = function(options = {}) {
```
stephenyeargin, line 237: This is a change in behavior to allow us to pass in override options instead of being constrained to only the `format` one.

symfony/src/Controller/DeceasedReportsController.php
```diff
@@ -0,0 +1,133 @@
...
+        return $this->render('deceasedreports/new.html.twig', [
+            'participant' => $participant,
+            'report' => $report,
+            'form' => $form->createView()
+        ]);
+    }
+
+    /* Private Methods */
+
+    public function formatReportTableRows($reports = [])
```
jt2k, line 108: Why are these not actually private?

symfony/src/Entity/PatientStatus.php
```diff
@@ -1,134 +0,0 @@
-<?php
```
jt2k, line 1: Why is this removed?

symfony/src/Entity/PatientStatusHistory.php
```diff
@@ -1,188 +0,0 @@
-<?php
```
jt2k, line 1: Why is this removed?

symfony/src/Entity/PatientStatusImport.php
```diff
@@ -1,243 +0,0 @@
-<?php
```
jt2k, line 1: Why is this removed?

jt2k (commented):
Still working through this, but had a few immediate questions.

symfony/src/Entity/PatientStatus.php
stephenyeargin: I must have picked up the delete when trying to clean up after running `doctrine:mapping:import`. 🐼 Fixed in 
9e6d25f

symfony/src/Entity/PatientStatusHistory.php
stephenyeargin: Resolved in 9e6d25f

symfony/src/Entity/PatientStatusImport.php
stephenyeargin: Resolved in 9e6d25f

symfony/src/Controller/DeceasedReportsController.php
stephenyeargin: Should have been. The other method was moved entirely, so it's no longer used. Resolved in c672ddc

symfony/src/Controller/DeceasedReportsController.php
```diff
@@ -0,0 +1,119 @@
...
+ * @Route("/s/deceased-reports")
+ */
+class DeceasedReportsController extends AbstractController
+{
+    /**
+     * @Route("/", name="deceased_reports_index")
+     */
+    public function participantObservationIndex(Request $request, SessionInterface $session, DeceasedReportsService $deceasedReportsService) {
+        $statusFilter = $request->query->get('status', 'preliminary');
+        $organizationId = ($request->query->get('all', false) === false) ? $session->get('siteOrganizationId', null) : null;
```
jt2k, line 24: Should we allow for getting all reports? My thought here is that if the user doesn't have an organization associated with their current site for some reason, we should redirect them out of here.

symfony/src/Service/DeceasedReportsService.php
```diff
@@ -0,0 +1,152 @@
...
+class DeceasedReportsService
+{
+    protected $api;
+
+    public function __construct(RdrApiService $api)
+    {
+        $this->api = $api;
+    }
+
+    public function getDeceasedReports($organizationId = null, $status = null)
```
jt2k, line 18: Related to comment above - maybe `$organizationId` shouldn't be allowed to be null.

symfony/src/Controller/DeceasedReportsController.php
```diff
@@ -0,0 +1,119 @@
...
+        if ($form->isSubmitted() && $form->isValid()) {
+            $report = $form->getData();
+            $form = $this->createForm(DeceasedReportType::class, $report, ['disabled' => true]);
+            try {
+                $fhirData = $deceasedReportsService->buildDeceasedReportFhir($report, $this->getUser());
+                $response = $deceasedReportsService->createDeceasedReport($participantId, $fhirData);
+                $report = $report->loadFromFhirObservation($response);
+                $this->addFlash('success', 'Deceased Report created!');
+            } catch (\Exception $e) {
+                $this->addFlash('error', $e->getMessage());
```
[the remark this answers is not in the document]
stephenyeargin: I had cleaned this up on L57, but not here for the error message. Ideally whatever error you encountered would be caught with validation.

The read-only setting is a product of the state of the submission in the event of an error. I *think* I can drop L88 and be fine as long as I set the form itself back to `preliminary`.

symfony/src/Controller/DeceasedReportsController.php
```diff
@@ -0,0 +1,119 @@
...
+     * @Route("/{participantId}/{reportId}", name="deceased_report_review", requirements={"participantId"="P\d+","reportId"="\d+"})
+     */
+    public function deceasedReportReview(Request $request, ParticipantSummaryService $participantSummaryService, DeceasedReportsService $deceasedReportsService, $participantId, $reportId) {
+        $participant = $participantSummaryService->getParticipantById($participantId);
+        if (!$participant) {
+            throw $this->createNotFoundException('Participant not found.');
+        }
+        $fhirData = $deceasedReportsService->getDeceasedReportById($participantId, $reportId);
+        $report = (new DeceasedReport())->loadFromFhirObservation($fhirData);
+        if ($report->getSubmittedBy() == $this->getUser()->getEmail() && $report->getReportStatus() == 'preliminary') {
```
jt2k, line 43: Can we change these (and several other `==` comparisons throughout the PHP code) to strict comparisons?

symfony/src/Controller/DeceasedReportsController.php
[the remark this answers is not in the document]
stephenyeargin: Or maybe I didn't clean it up. Trouble there from the submission and submission review being the same object.

symfony/src/Controller/DeceasedReportsController.php
stephenyeargin: Fixed in 11f84c8

symfony/src/Service/DeceasedReportsService.php
stephenyeargin: I'm hesitant to do this as the service _might_ be used to grab all `DeceasedReports` for the email notification bit (depending on if we process those by Organization or if we collect them all the pending approval ones and then only get the organization details for the batch). If it's no longer optional, I'd want to rename the method to `getDeceasedReportsByOrganizationId()`

symfony/src/Controller/DeceasedReportsController.php
stephenyeargin: This was mostly for my debugging. Dropped.

symfony/src/Service/EnvironmentService.php
stephenyeargin: Seems to work as expected on the Test server (after some adjustments in 02168cc).

symfony/src/Controller/DeceasedReportsController.php
```diff
@@ -0,0 +1,124 @@
...
+        }
+        $form = $this->createForm(DeceasedReportType::class, $report, ['disabled' => (bool) $report->getId()]);
+        $form->handleRequest($request);
+        if ($form->isSubmitted() && $form->isValid()) {
+            $report = $form->getData();
+            try {
+                $fhirData = $deceasedReportsService->buildDeceasedReportFhir($report, $this->getUser());
+                $response = $deceasedReportsService->createDeceasedReport($participantId, $fhirData);
+                $report = $report->loadFromFhirObservation($response);
+                $this->addFlash('success', 'Deceased Report created!');
```
shyamnaru, line 96: It looks like the form fields are still editable after the report as been created. We may need to redirect back to this route to make those fields disabled?
`return $this->redirectToRoute('deceased_report_new', ['participantId' => $participantId]);`

jt2k: Good point. It is also confusing because you get the success message and also the info alert that a report has already been created. Does the ticket or PRD say where the user should go after submitting the report? If not, I'm thinking it probably makes sense to redirect back to the participant summary page, since that is most likely where the user came from.

stephenyeargin: Ah, that's the reason the "disabled" bit was in there earlier.

Redirecting to participant summary page is ideal, but we may have interesting logic to add with the discussion today about being able to initiate the process from different account types. For now, we can send it to `participant` route with the success flash message.

stephenyeargin: Resolved in e4fe099f

jt2k (approved):
🎉

shyamnaru (approved):

stephenyeargin:
Merging this into `feature/deceased-participants` so we can wrap in the other requirements.

PR #730, opened 2020-09-11
[HPRO-683] Deceased Reports > Add badge to Welcome Page

stephenyeargin:
Screen Shot 2020-09-11 at 5 02 53 PM

Adds a badge that shows the number of Pending Deceased Reports for the logged in user on the main Welcome page.

Note: It takes a second or so for it to show up, and will only appear if the organization has pending reports (i.e. it won't appear if there are zero pending reports).

[HPRO-683]

[HPRO-683]: https://precisionmedicineinitiative.atlassian.net/browse/HPRO-683

### DIFF
--- a/symfony/src/Controller/DeceasedReportsController.php
+++ b/symfony/src/Controller/DeceasedReportsController.php
@@ -6,10 +6,12 @@ use App\Form\DeceasedReportReviewType;
 use App\Form\DeceasedReportType;
 use App\Service\DeceasedReportsService;
 use App\Service\ParticipantSummaryService;
+use Pmi\Cache\DatastoreAdapter;
 use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Session\SessionInterface;
 use Symfony\Component\Routing\Annotation\Route;
+use Symfony\Contracts\Cache\ItemInterface;
 
 /**
  * @Route("/s/deceased-reports")
@@ -106,6 +108,30 @@ class DeceasedReportsController extends AbstractController
             'participant' => $participant,
             'report' => $report,
             'form' => $form->createView()
+        ]);
+    }
+
+    /**
+     * @Route("/stats", name="deceased_report_stats")
+     */
+    public function getStats(SessionInterface $session, DeceasedReportsService $deceasedReportsService)
+    {
+        $organizationId = $session->get('siteOrganizationId');
+        $cache = new DatastoreAdapter(100);
+        $pendingReportCountCache = $cache->getItem('deceased_reports.pending.count');
+        if ($cacheHit = $pendingReportCountCache->isHit()) {
+            $pendingReportCount = (int) $pendingReportCountCache->get();
+        } else {
+            $pendingReportCountCache->expiresAfter(60);
+            $reports = $deceasedReportsService->getDeceasedReports($organizationId, 'preliminary');
+            $pendingReportCount = count($reports);
+            $pendingReportCountCache->set($pendingReportCount);
+            $cache->save($pendingReportCountCache);
+        }
+
+        return $this->json([
+            'pending' => $pendingReportCount,
+            'cacheHit' => $cacheHit
         ]);
     }
 
--- a/symfony/src/Controller/DeceasedReportsController.php
+++ b/symfony/src/Controller/DeceasedReportsController.php
@@ -18,6 +18,9 @@ use Symfony\Contracts\Cache\ItemInterface;
  */
 class DeceasedReportsController extends AbstractController
 {
+    const ORG_PENDING_CACHE_TTL = 500;
+    const ORG_PENDING_CACHE_KEY = 'deceased_reports.org-%s.pending.count';
+
     /**
      * @Route("/", name="deceased_reports_index")
      */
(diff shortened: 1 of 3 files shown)


symfony/src/Controller/DeceasedReportsController.php
```diff
@@ -109,6 +111,30 @@ public function deceasedReportNew(Request $request, ParticipantSummaryService $p
...
+     */
+    public function getStats(SessionInterface $session, DeceasedReportsService $deceasedReportsService)
+    {
+        $organizationId = $session->get('siteOrganizationId');
+        $cache = new DatastoreAdapter(100);
+        $pendingReportCountCache = $cache->getItem('deceased_reports.pending.count');
+        if ($cacheHit = $pendingReportCountCache->isHit()) {
+            $pendingReportCount = (int) $pendingReportCountCache->get();
+        } else {
+            $pendingReportCountCache->expiresAfter(60);
```
stephenyeargin, line 125: Can increase this to a higher number, but since reports can come from platforms other than HealthPro, we could wind up with an inaccurate count.

views/index.html.twig
```diff
@@ -81,3 +81,20 @@
...
     </div>
 {% endblock %}
+
+{% block pagejs %}
+<script>
+$(document).ready(function () {
+  // Add badge for count of Deceased Participant Reports for review
+  var deceasedReportCount = 0;
+  $.ajax({
+    url: '/s/deceased-reports/stats'
```
stephenyeargin, line 91: Hard-coding this for now. I use [FOSJsRoutingBundle](https://symfony.com/doc/master/bundles/FOSJsRoutingBundle/usage.html) in other projects, but this page has not been migrated over.

symfony/src/Controller/DeceasedReportsController.php
```diff
@@ -109,6 +111,30 @@ public function deceasedReportNew(Request $request, ParticipantSummaryService $p
...
+            $reports = $deceasedReportsService->getDeceasedReports($organizationId, 'preliminary');
+            $pendingReportCount = count($reports);
+            $pendingReportCountCache->set($pendingReportCount);
+            $cache->save($pendingReportCountCache);
+        }
+
+        return $this->json([
+            'pending' => $pendingReportCount,
+            'cacheHit' => $cacheHit
```
shyamnaru, line 134: Where are we using this?

stephenyeargin:
Loaded from https://github.com/all-of-us/healthpro/blob/0e68925e64e31a777c6ec40c88f0f417b28e3388/views/index.html.twig#L91

symfony/src/Controller/DeceasedReportsController.php
stephenyeargin: This was mostly for debugging purposes. I typically pass along something like this when working with a cached endpoint if I need to demonstrate that a cache "miss" is happening.

shyamnaru (approved):

views/index.html.twig
jt2k: Currently, the way we are passing routes to JS is through the `PMI.path` global object (see `base-head.html.twig` in both the Symfony and Silex templates). It's not necessarily the best solution, but just thought I'd point that out in case you think it makes sense to use.

symfony/src/Controller/DeceasedReportsController.php
jt2k: I'm thinking at least 5 minutes would be better. Otherwise, that's a lot of requests off to the RDR every time the home page is loaded.

We could consider some more complex caching logic, too. Maybe set the TTL longer if the count is 0 and shorter if it is non-0. And we could clear the organization's cache value whenever a deceased report is submitted or reviewed, since we know for sure it will change at that point.

symfony/src/Controller/DeceasedReportsController.php
```diff
@@ -109,6 +111,30 @@ public function deceasedReportNew(Request $request, ParticipantSummaryService $p
...
     }
 
+    /**
+     * @Route("/stats", name="deceased_report_stats")
+     */
+    public function getStats(SessionInterface $session, DeceasedReportsService $deceasedReportsService)
+    {
+        $organizationId = $session->get('siteOrganizationId');
+        $cache = new DatastoreAdapter(100);
+        $pendingReportCountCache = $cache->getItem('deceased_reports.pending.count');
```
jt2k, line 121: Shouldn't the cache key include the organization id?

symfony/src/Controller/DeceasedReportsController.php
stephenyeargin: Will up it to five minutes. We can clear _our_ version of the cache, but would we have overlap with PMT users?

symfony/src/Controller/DeceasedReportsController.php
stephenyeargin: Uh, yes. 😬

symfony/src/Controller/DeceasedReportsController.php
jt2k: We can take this back to Caroline, but IMO, 5 minutes is an acceptable delay. And if we do some optimization based on the stuff we do know, even better.

views/index.html.twig
stephenyeargin: Thanks. If we started building it into the the overall "chrome" of the app (global alerts) I'd feel better about putting it there. For now, I think I'll leave it as the hard-coded URL since it is referenced from only the one page.

symfony/src/Controller/DeceasedReportsController.php
stephenyeargin: Addressed in 3d7a541

symfony/src/Controller/DeceasedReportsController.php
stephenyeargin: Addressed in 3d7a541. Can bring it up in tomorrow's meeting to see if it's a major concern.

views/index.html.twig
```diff
@@ -81,3 +81,20 @@
...
+  }).done(function (data) {
+    deceasedReportCount = data.pending
+    if (deceasedReportCount > 0) {
+      $('#deceased_report_block').prepend($('<div class="welcome-page-badge">' + deceasedReportCount +'</div>'));
```
jt2k, line 95: This causes a problem with the `equalizeColumnHeight` plugin in `app.js`. The purpose of this plugin is to make all of the contents across columns the same height. The plugin assumes that the immediate children of the column divs should be equalized (the btn-block elements in this case). Is it possible to make this work where the badge is inside the `a.btn-block` instead of the parent column div?

Here's what currently happens when you resize the window after the badge is loaded:

Screen Shot 2020-09-14 at 3 18 59 PM

stephenyeargin: Resolved in ed8c73e. Has the added bonus of the badge being clickable as well. Needed to adjust the font size down a smidge.

jt2k (approved):

PR #735, opened 2020-09-18
[HPRO-679] Deceased Reports > Email Notifications

shyamnaru:


symfony/src/Service/DeceasedNotificationService.php
```diff
@@ -0,0 +1,76 @@
...
+        EntityManagerInterface $em,
+        ManagerRegistry $managerRegistry,
+        ParticipantSummaryService $participantSummaryService,
+        LoggerService $loggerService,
+        EnvironmentService $env,
+        ParameterBagInterface $params,
+        Environment $twig,
+        SiteRepository $siteRepository,
+        DeceasedLogRepository $deceasedLogRepository
+    ) {
```
jt2k, line 42: Can we reduce the number of injected dependencies? Since you need already have the entity manager, you could avoid injecting the specific entity repositories and just use the entity manager to get the required repositories (e.g. `$this->em->getRepository(Site::class)`)

It would also be ideal to just need one of an `EntityManagerInterface` or `ManagerRegistry`. I haven't tested this, but it looks like the `ManagerRegistry` has a `getManager` method which might return the entity manager. In that case, you could just inject the manager registry and then set `$this->em = $managerRegistry->getManager()`.

symfony/src/Entity/DeceasedLog.php
```diff
@@ -0,0 +1,125 @@
...
+
+    /**
+     * @ORM\Column(type="datetime", nullable=true)
+     */
+    private $deceasedTs;
+
+    /**
+     * @ORM\Column(type="string", length=100, nullable=true)
+     */
+    private $hpoId;
```
jt2k, line 37: hpoId is a confusing column name. Don't we have places where hpoId is used as a legacy synonym for awardee ID? Shouldn't this be representing an organization ID instead?

symfony/src/Service/DeceasedNotificationService.php
```diff
@@ -0,0 +1,76 @@
...
+        $this->deceasedStatus = $deceasedStatus;
+        $this->render = $this->statusText = "deceased-{$deceasedStatus}";
+        $this->log = $deceasedStatus === 'pending' ? Log::DECEASED_PENDING_NOTIFY : Log::DECEASED_APPROVED_NOTIFY;
+    }
+
+    protected function getSearchParams($id, $lastDeceased)
+    {
+        $searchParams = [
+            'deceasedStatus' => strtoupper($this->deceasedStatus),
+            'hpoId' => $id,
```
jt2k, line 65: I made a comment above about hpoId, but now I realize I have broader questions. Shouldn't this search be by organization ID? I thought we are looking for deceased participants and notifying at an organization level.

symfony/src/EventListener/RequestListener.php
```diff
@@ -73,7 +73,7 @@ private function checkPageNotices()
 
     private function checkSiteSelect()
     {
-        if (!$this->session->has('site')) {
+        if (!$this->session->has('site') && !preg_match('/^\/s\/cron\/.*/', $this->request->getPathInfo())) {
```
jt2k, line 76: Can we make the `/s` prefix optional like in the regex above for the admin URL pattern regex?

symfony/src/Security/GoogleGroupsAuthenticator.php
```diff
@@ -28,7 +28,9 @@ public function __construct(UserService $userService, ContainerBagInterface $par
 
     public function supports(Request $request)
     {
-        if ($request->getSession()->has('isLogin') && $request->getSession()->has('_security_main')) {
+        if (($request->getSession()->has('isLogin') && $request->getSession()->has('_security_main')) ||
+            (preg_match('/^\/s\/cron\/.*/', $request->getPathInfo()) && ($request->headers->get('X-Appengine-Cron') === 'true' || $this->env->isLocal()))
```
jt2k, line 32: Same as in the listener - can we make the `/s` prefix optional?

symfony/src/Service/DeceasedNotificationService.php
```diff
@@ -30,25 +29,22 @@ class DeceasedNotificationService extends EmailNotificationService
...
         $this->managerRegistry = $managerRegistry;
         $this->participantSummaryService = $participantSummaryService;
         $this->loggerService = $loggerService;
         $this->env = $env;
         $this->params = $params;
         $this->twig = $twig;
-        $this->siteRepository = $siteRepository;
-        $this->logRepository = $deceasedLogRepository;
+        $this->siteRepository = $this->em->getRepository(Site::class);
+        $this->logRepository = $this->em->getRepository(DeceasedLog::class);
```
jt2k: Thanks for cleaning this up. I just noticed that this class redefines the same class variables as in the parent EmailNotificationService class. Can you remove the ones that aren't set here? You should only need to include the ones that aren't in the parent class. Or are overriding the default value.

jt2k (approved):

stephenyeargin (approved):
📬

PR #743, opened 2020-09-23
[HPRO-630] Redact submission when Deceased Report is denied

stephenyeargin:
Screen Shot 2020-09-23 at 1 14 46 PM

No need to show the previously submitted information once reviewed as denied.

[HPRO-630]

[HPRO-630]: https://precisionmedicineinitiative.atlassian.net/browse/HPRO-630

### DIFF
--- a/symfony/templates/deceasedreports/review.html.twig
+++ b/symfony/templates/deceasedreports/review.html.twig
@@ -18,7 +18,8 @@
       {% if report.denialReason %}
       <dt>Denial Reason</dt>
       <dd>{{ report.denialReasonDisplay }} {% if report.denialReason == 'OTHER' %}- {{ report.denialReasonOtherDescription }}{% endif %}</dd>
-      {% endif %}
+
+      {% else %}
 
       <hr />
       <dt>Date of Death</dt>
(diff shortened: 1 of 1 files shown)


symfony/templates/deceasedreports/review.html.twig
```diff
@@ -18,7 +18,8 @@
       {% if report.denialReason %}
       <dt>Denial Reason</dt>
       <dd>{{ report.denialReasonDisplay }} {% if report.denialReason == 'OTHER' %}- {{ report.denialReasonOtherDescription }}{% endif %}</dd>
-      {% endif %}
+
+      {% else %}
```
jt2k, line 22: Should we account for `reportStatus` = `cancelled` but no denial reason is set? This should be required, but seems like it would be good to be explicit here. That is, display the denial reason if it is set. And then if the status is not denied, show the information currently in this else block.

stephenyeargin: Not a bad move. I'm _pretty_ sure creating a review in the RDR explicitly requires the reason (fails on creation without it), but that would be an extra safeguard. Fixed in f74c2b51

jt2k (approved):

shyamnaru (approved):

PR #744, opened 2020-09-23
[HPRO-687] Disable deceased reporting for withdrawn/deactivated participants

stephenyeargin:
This PR removes the cog icon on the Participant Summary page for deactivated and withdrawn participants. It also adds logic that prevents new Deceased Reports from being created for those participants.

Additionally, it fixes a bug where the Deactivated filter would not work in the Work Queue because of a restriction on combining it with Deceased Status.

[HPRO-687]

[HPRO-687]: https://precisionmedicineinitiative.atlassian.net/browse/HPRO-687


stephenyeargin:
Side note: I'm realizing that our `ParticipantSummaryService` should likely return a hydrated `Participant` object instead of the raw RDR output so that we can leverage logic already present in that class.

jt2k (approved):

shyamnaru (approved):

PR #747, opened 2020-09-25
[HPRO-669] Add "Return to Participant" button for finalized Biospecimen orders

shyamnaru:


stephenyeargin (approved):

views/order-header.html.twig
```diff
@@ -15,6 +15,9 @@
     {% set params = { participantId: participant.id, orderId: order.id } %}
     {% set route = app.request_stack.currentRequest.attributes.get('_route') %}
     {% set return = path(route, params) %}
+    {% if order and order.finalized_ts and order.rdr_id %}
+        <a href="{{ path('participant', { id: participant.id }) }}" class="btn btn-default">Return to participant</a>
```
jt2k, line 19: Should we add the `btn-sm` class to match the other buttons?

jt2k (commented):
I left a comment about the button size, but more generally, as I was looking at this, I'm wondering if we should review this with Caroline. The intent of the ticket was to make this consistent with the physical measurements form, but this seems pretty different to me. Especially since a link back to the participant is on the participant's name in the header right above this new button.

I will add a comment and screenshot to discuss on the JIRA ticket.

shyamnaru:
@jt2k I was thinking the same. Actually, I added the button at the bottom of the form then saw the ticket description and changed it to top. Good to review this with Caroline.

shyamnaru:
Closing this PR as we decided not to implement this.

PR #748, opened 2020-09-25
[HPRO-662] Add Oct COPE survey details to the Participant Work Queue and the Participant Summary pages

shyamnaru:


stephenyeargin (approved):

jt2k (approved):
Looks good. We should probably wait until we get confirmation that this is going to be ready in RDR prod with correct data before merging.

PR #750, opened 2020-10-05
[HPRO-692] Fix patient status bulk import issue

shyamnaru:
This PR fixes combination of duplicate participant_id and organization insertions in the patient_status table

### DIFF
--- a/src/Pmi/Controller/DefaultController.php
+++ b/src/Pmi/Controller/DefaultController.php
@@ -447,15 +447,15 @@ class DefaultController extends AbstractController
             $patientStatusForm = $patientStatus->getForm($isCommentRequired);
             $patientStatusForm->handleRequest($request);
             if ($patientStatusForm->isSubmitted()) {
-                $patientStatusData = $app['em']->getRepository('patient_status')->fetchOneBy([
+                $patientStatusData = $app['em']->getRepository('patient_status')->fetchBy([
                     'participant_id' => $id,
                     'organization' => $app->getSiteOrganizationId()
-                ]);
+                ], ['history_id' => 'desc'], 1);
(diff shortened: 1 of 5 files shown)


stephenyeargin (approved):

jt2k (commented):
Should we also add a check to the import upload? We could make it so that uploads with multiple rows with the same participant ID are treated as invalid and not processed.

shyamnaru:
Yes, we could throw a form error when the import file contains duplicate participant ids.

src/Pmi/Controller/DefaultController.php
```diff
@@ -447,15 +447,15 @@ public function participantAction($id, Application $app, Request $request)
...
                     'participant_id' => $id,
                     'organization' => $app->getSiteOrganizationId()
-                ]);
+                ], ['history_id' => 'desc'], 1);
                 if (!empty($patientStatusData) && empty($patientStatusForm['comments']->getData())) {
                     $patientStatusForm['comments']->addError(new FormError('Please enter comment'));
                 }
                 if ($patientStatusForm->isValid()) {
-                    $patientStatusId = !empty($patientStatusData) ? $patientStatusData['id'] : null;
+                    $patientStatusId = !empty($patientStatusData) ? $patientStatusData[0]['id'] : null;
```
jt2k, line 458: Now that we're adding the unique constraint and cleaning up the data, do we still need these changes?

shyamnaru: Yes, we don't need them, I'll update. Thanks!

stephenyeargin (approved):

jt2k (approved):

PR #756, opened 2020-10-15
[HPRO-665] Migrate admin site management to symfony

shyamnaru:


symfony/src/Entity/Site.php
```diff
@@ -90,7 +90,7 @@ class Site
     /**
      * @ORM\Column(type="smallint")
```
jt2k, line 91: This was already here before, but since we're updating the Site entity, should we change this to a boolean type? You'll need to check where else we are using this and make any adjustments if necessary.

jt2k: The status column should probably be a boolean also.

symfony/src/Entity/Site.php
```diff
@@ -276,4 +276,9 @@ public function setDeleted(int $deleted): self
 
         return $this;
     }
+
+    public function toArray()
+    {
+        return get_object_vars($this);
+    }
```
jt2k, line 283: While in the current case, this works fine, I think it would be a better practice to not automatically expose all of the object variables like this. We should consider using the [Serializer component](https://symfony.com/doc/4.4/components/serializer.html). (In the current case where this is used for the site sync diff, you would only need to normalize, and not serialize).

symfony/src/Service/SiteSyncService.php
```diff
@@ -0,0 +1,257 @@
...
+            }
+        }
+        if (empty($organizationsMap)) {
+            throw new \Exception('No organizations found');
+        }
+        $cmd = $this->em->getClassMetadata(Organization::class);
+        $connection = $this->em->getConnection();
+        $connection->beginTransaction();
+        try {
+            $connection->query('DELETE FROM ' . $cmd->getTableName());
```
jt2k, line 242: Can you use the query builder or DQL to perform this delete instead of using raw SQL?

symfony/src/Service/SiteSyncService.php
```diff
@@ -0,0 +1,257 @@
...
+            $awardeesMap[trim($awardee->id)] = $awardee->displayName;
+        }
+        if (empty($awardeesMap)) {
+            throw new \Exception('No awardees found');
+        }
+        $cmd = $this->em->getClassMetadata(Awardee::class);
+        $connection = $this->em->getConnection();
+        $connection->beginTransaction();
+        try {
+            $connection->query('DELETE FROM ' . $cmd->getTableName());
```
jt2k, line 203: Can you use the query builder or DQL to perform this delete instead of using raw SQL?

symfony/src/Service/SiteSyncService.php
```diff
@@ -0,0 +1,260 @@
...
+
+    public function sync($preview = false)
+    {
+        $sitesCount = 0;
+        $created = [];
+        $modified = [];
+        $existingSites = $this->getSitesFromDb();
+        $deleted = array_keys($existingSites); // add everything to the deleted array, then remove as we find them
+        $entries = $this->getAwardeeEntriesFromRdr();
+        $serializer = new Serializer([new ObjectNormalizer()]);
```
jt2k, line 80: Instead of instantiating a new serializer and having to pass in a normalizer, you can just inject the normalizer into the service constructor:

```php
use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
...
// add $normalizer class variable, add `NormalizerInterface $normalizer` to constructor parameters and assign
...
// then, use it where needed:
$this->normalizer->normalize($existing);
...
```
I tested this and the result is the same as the ObjectNormalizer.

jt2k (approved):
🎉

stephenyeargin (approved):
:shipit:

stephenyeargin (approved):

PR #763, opened 2020-10-22
[HPRO-667] Migrate admin debug participant lookup tool to symfony

shyamnaru:


symfony/src/Controller/DebugToolsController.php
```diff
@@ -0,0 +1,57 @@
...
+ */
+class DebugToolsController extends AbstractController
+{
+    /**
+     * @Route("/participants", name="admin_debug_participants")
+     */
+    public function participantsAction(Request $request, EnvironmentService $env, DebugToolsService $debugToolsService)
+    {
+        if ($env->isProd()) {
+            throw $this->createNotFoundException('Page Not Found!');
```
jt2k, line 23: If we don't have a need to customize the 404 exception, you can just leave the constructor empty. (Same in the other calls below).

stephenyeargin (approved):

jt2k (approved):

PR #771, opened 2020-10-27
[HPRO-663] Invalid HTML in panel body for blood pressure average on the Physical Mesaurements form

shyamnaru:


web/assets/css/app.css
```diff
@@ -442,3 +442,9 @@ table.table-samples td {
   border: solid 2pt white;
   box-shadow: 0 .5rem 1rem rgba(0,0,0,.15);
 }
+
+.bp-warnings .alert {
```
jt2k, line 446: Can you move this up to where the other styles for physical measurements are (and maybe add the `#physicalEvaluation` div selector in front too).

jt2k (approved):

PR #777, opened 2020-11-12
[HPRO-721] Refactor Participant::getActivityStatus

stephenyeargin:
The hierarchy for activity status has been reorganized, and as a result this PR refactors the logic for returning the correct string.

The priority for these should be:

1. Withdrawn
1. Deactivated
1. Deceased

[HPRO-721]

[HPRO-721]: https://precisionmedicineinitiative.atlassian.net/browse/HPRO-721

### DIFF
--- a/src/Pmi/Entities/Participant.php
+++ b/src/Pmi/Entities/Participant.php
@@ -383,20 +383,28 @@ class Participant
 
     private function getActivityStatus($participant)
     {
+        // Withdrawn
         if (in_array($participant->withdrawalStatus, self::$withdrawalStatusValues, true)) {
             return 'withdrawn';
-        } elseif (in_array($participant->deceasedStatus, self::$deceasedStatusValues, true)) {
-            return 'deceased';
-        } else {
-            switch (isset($participant->suspensionStatus) ? $participant->suspensionStatus : null) {
+        }
+
+        // Deactivated
+        if (isset($participant->suspensionStatus)) {
+            switch ($participant->suspensionStatus) {
                 case 'NOT_SUSPENDED':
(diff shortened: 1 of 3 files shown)


src/Pmi/Entities/Participant.php
```diff
@@ -383,20 +383,28 @@ private function getSiteSuffix($site)
...
-            return 'deceased';
-        } else {
-            switch (isset($participant->suspensionStatus) ? $participant->suspensionStatus : null) {
+        }
+
+        // Deactivated
+        if (isset($participant->suspensionStatus)) {
+            switch ($participant->suspensionStatus) {
                 case 'NOT_SUSPENDED':
                     return 'active';
```
jt2k, line 395: I think in the case of suspensionStatus = NOT_SUSPENDED, we then want to check for the deceased status. I think suspension status should always be NO_CONTACT or NOT_SUSPENDED. We can confirm with the RDR.

I believe what we want is just:

```php
if (isset($participant->suspensionStatus) && $participant->suspensionStatus === 'NO_CONTACT')) {
    return 'deactivated';
}
```
 Then, check for deceased. And then return active if none apply.

We should check to see if an empty value should ever be returned.

stephenyeargin: 🐼 I'd forgotten that these were not mutually exclusive statuses.

stephenyeargin: Resolved in f75a22d3

jt2k (approved):
Looks good. Thanks for adding the tests!

shyamnaru (approved):

PR #786, opened 2020-11-25
[HPRO-717] Enable SDBB Diversion Pouch DV Sites to do HPO Biospecimen Collection

shyamnaru:


stephenyeargin:
Actually, looks like `google_analytics_property` is already there, but commented out by default. This change made it mandatory in order to view the `/help` route.

shyamnaru:
> google_analytics_property

shyamnaru:
Sorry, I accidentally closed and had to reopen it.

dev_config/config.yml.dist
```diff
@@ -67,7 +67,8 @@ enforce2fa: 0
 sites_use_rdr: true
 
 # Enable Google Analytics by setting property ID
-# google_analytics_property: UA-XXXXXXXX-NN
+google_analytics_property: UA-XXXXXXXX-NN
```
jt2k, line 70: This was changed in response to Stephen's comment. But I think the better change would be to make the Twig globals not break if these optional configuration items are not configured.

symfony/config/packages/twig.php
```diff
@@ -8,11 +8,10 @@
 $container->loadFromExtension('twig', [
     'globals' => [
         'isStable' => $env->isStable(),
-        'isDvType' => '',
-        'reportKitUrl' => '',
+        'reportKitUrl' => $env->configuration['reportKitUrl'],
         'assetVer' => $env->values['assetVer'],
         'awardeeOrganization' => '',
-        'google_analytics_property' => '',
+        'google_analytics_property' => $env->configuration['google_analytics_property'],
```
jt2k, line 14: As mentioned above in the config file, I think this is where we need to make an adjustment to use an empty value if a non-critical configuration item like these isn't set, instead of breaking due to an unset array key (I'm assuming that's the reason this breaks).

jt2k (approved):
🎉

stephenyeargin (approved):

PR #816, opened 2021-01-26
[HPRO-748] Migrate order workflow to Symfony

shyamnaru:
| Q                   | A
| Bug fix?            | ❌ 
| New feature?        | ❌ 
| Database migration? |❌ 
| New configuration?  | ❌ 
| Composer updates?   | ❌ 
| NPM updates?        | ❌ 
| Jira Ticket(s)      | HPRO-748


symfony/src/Service/ParticipantSummaryService.php
```diff
@@ -19,7 +20,8 @@ public function getParticipantById($participantId)
     {
         try {
             $response = $this->api->get(sprintf('rdr/v1/Participant/%s/Summary', $participantId));
-            return json_decode($response->getBody());
+            $participant = json_decode($response->getBody()->getContents());
+            return new Participant($participant);
```
stephenyeargin, line 24: Quick note here to make sure we test other implementations using `ParticipantSummaryService::getParticipantById()` (thinking mostly of deceased reports), as I believe those might be expecting the JSON object instead of the entity. In hindsight, it was my mistake not to do it this way to begin with.

shyamnaru: Thanks for checking. Actually the `Participant` class just returns the same JSON object structure with some additional properties attached to it. I noticed this before and made sure it didn't break other implementations.

stephenyeargin:
Noticed this when merging it back into my work.

```
root@healthpro-web:/app# bin/symfony doctrine:schema:validate

Mapping
-------

 [FAIL] The entity-class App\Entity\Measurement mapping is invalid:
 * The field App\Entity\Measurement#history is on the owning side of a bi-directional relationship, but the specified mappedBy association on the target-entity App\Entity\MeasurementHistory# does not contain the required 'inversedBy' attribute.


 [FAIL] The entity-class App\Entity\MeasurementHistory mapping is invalid:
 * The association App\Entity\MeasurementHistory#measurement refers to the inverse side field App\Entity\Measurement#measurementHistory which does not exist.


Database
--------

                                                                                                                        
 [ERROR] The database schema is not in sync with the current mapping file.                                              
                                                                                                                        
```

symfony/config/packages/twig.php
```diff
@@ -15,6 +15,7 @@
         'sessionTimeout' => $env->values['sessionTimeOut'],
         'sessionWarning' => $env->values['sessionWarning'],
         'timeZones' => $env->getTimeZones(),
-        'confluenceResources' => HelpService::$confluenceResources
+        'confluenceResources' => HelpService::$confluenceResources,
+        'orderSampleInfo' => $env->configuration['order_samples_info_3']
```
jt2k, line 19: This currently breaks if this configuration item isn't set. Can we add a null coalesce here? Something like:

```suggestion
        'orderSampleInfo' => $env->configuration['order_samples_info_3'] ?? null
```

symfony/src/Service/OrderService.php
```diff
@@ -2,23 +2,54 @@
...
+        UserService $userService,
+        SiteService $siteService,
+        LoggerService $loggerService
+    ) {
         $this->rdrApiService = $rdrApiService;
         $this->params = $params;
+        $this->em = $em;
+        $this->mayolinkOrderService = $mayolinkOrderService;
+        $this->userService = $userService;
+        $this->mayolinkOrderService = $mayolinkOrderService;
```
jt2k, line 43: This is getting set twice

symfony/src/Entity/Order.php
```diff
@@ -644,14 +671,44 @@ public function setFinalizedSite(?string $finalizedSite): self
...
+        return $this->currentVersion;
+    }
+
     public function loadSamplesSchema($params = [])
     {
-        $currentVersion = $this->getVersion();
+        $this->currentVersion = $this->getVersion();
         if (!empty($params['order_samples_version'])) {
-            $currentVersion = $params['order_samples_version'];
+            $this->currentVersion = $params['order_samples_version'];
```
jt2k, line 708: Shouldn't the entity's version, if it exists, take precedence over the configured current version? That is, if you pull up an old order with a different version from the current version, we would want to load the sample information for the old version.

symfony/src/Service/MayolinkOrderService.php
```diff
@@ -0,0 +1,155 @@
...
+        $this->em = $em;
+        $this->twig = $twig;
+        $this->logger = $logger;
+        $this->client = new HttpClient(['cookies' => true]);
+        $this->ordersEndpoint = $params->get('ml_orders_endpoint');
+        $this->userName = $params->get('ml_username');
+        $this->password = $params->get('ml_password');
+        if (empty($this->ordersEndpoint) || empty($this->userName) || empty($this->password)) {
+            throw new \Exception('MayoLINK connection is not configured.');
+        }
```
jt2k, line 41: Previously, we could have `ml_mock_order` defined locally and not need to specify the endpoint and credentials. Can we reinstate this behavior? Currently, this service needs to be wired even if using the mock orders, and it fails when the endpoint and credentials aren't defined.

jt2k (approved):

stephenyeargin (approved):
Great work!

PR #822, opened 2021-02-10
[HPRO-753] Migrate 'Work Queue' to Symfony

shyamnaru:
| Q                   | A
| Bug fix?            | ❌ 
| New feature?        | ❌ 
| Database migration? | ❌ 
| New configuration?  | ❌ 
| Composer updates?   | ❌ 
| NPM updates?        | ❌ 
| Jira Ticket(s)      | HPRO-753


symfony/config/packages/twig.php
```diff
@@ -10,7 +10,7 @@
         'isStable' => $env->isStable(),
         'reportKitUrl' => $env->configuration['reportKitUrl'] ?? '',
         'assetVer' => $env->values['assetVer'],
-        'awardeeOrganization' => '',
+        'site' => '@App\Service\SiteService',
```
stephenyeargin, line 13: I'm using this in my branch, so I know it works. I don't _think_ we'll ever run the risk of a name collision (e.g. say we were iterating through a list of `sites` and called the iterator `site`, wiping out the service in the process).

symfony/src/EventListener/RequestListener.php
```diff
@@ -73,7 +73,8 @@ private function checkPageNotices()
 
     private function checkSiteSelect()
     {
-        if (!$this->session->has('site') && !preg_match('/^(\/s)?\/(_profiler|_wdt|cron|admin|help|settings|problem|biobank|review)($|\/).*/', $this->request->getPathInfo())) {
+        if (!$this->session->has('site') && !preg_match('/^(\/s)?\/(_profiler|_wdt|cron|admin|help|settings|problem|biobank|review|workqueue)($|\/).*/',
```
stephenyeargin, line 76: I added the `review` one recently, but I can't remember -- should this be accessible if a site isn't set?

symfony/src/Twig/AppExtension.php
```diff
@@ -35,9 +38,9 @@ public function getFunctions()
             new TwigFunction('slugify', [$this, 'slugify']),
             new TwigFunction('timezone_display', [$this, 'timezoneDisplay']),
             new TwigFunction('codebook_display', [$this, 'getCodeBookDisplay']),
-            new TwigFunction('organization_display', [$this, 'getAwardeeDisplay']),
             new TwigFunction('awardee_display', [$this, 'getAwardeeDisplay']),
```
stephenyeargin, line 41: I think we can flag a few of these as deprecated since we're getting `SiteService` loaded in (help reduce redundant code).

symfony/src/EventListener/RequestListener.php
shyamnaru: Had to add this to enable super users (stsi) to access WorkQueue. Yes, this is not accessible if a site is not set. In the WorkQueue controller there is a check for site's awardee ([Line 66](https://github.com/all-of-us/healthpro/blob/sn/hpro-753/migrate-workqueue-to-symfony/symfony/src/Controller/WorkQueueController.php#L66)) and renders `no-organization.html.twig` file if it's empty (site is not set)

src/Pmi/Controller/SymfonyMigrationController.php
```diff
@@ -114,4 +115,13 @@ public function orderAction($participantId, $orderId, Application $app)
...
         ));
     }
+
+
+    /**
+     * @deprecated 2021-02-08
+     */
+    public function workqueue_indexAction(Application $app)
+    {
+        return $app->redirect('/s/workqueue');
```
jt2k, line 125: Missing the trailing slash. Symfony fixes this for you with the 301 redirect, but this breaks behind the WAF because Symfony uses an absolute URL for the redirect.

```suggestion
        return $app->redirect('/s/workqueue/');
```

symfony/src/Controller/WorkQueueController.php
```diff
@@ -0,0 +1,226 @@
...
+     * @var SiteService
+     */
+    protected $siteService;
+
+    /**
+     * WorkQueueController constructor.
+     * @param SessionInterface $session
+     * @param WorkQueueService $workQueueService
+     * @param SiteService $siteService
+     */
```
jt2k, line 39: I noticed in this PR you started introducing more docblocks with `@param` and `@return` lines. We can discuss more, but in my opinion, we should use native parameter types and return types instead.

symfony/src/Service/SiteService.php
```diff
@@ -2,38 +2,86 @@
...
             return $site->id;
-        } else {
-            return null;
         }
+        return null;
+    }
+
+    /**
+     * @return null
+     */
```
jt2k, line 73: I commented earlier about removing these in favor of return types, but wanted to also note that these two are incorrect.

jt2k (approved):

stephenyeargin (approved):

PR #841, opened 2021-03-24
[HPRO-772] [HPRO-771] Add EHR and Pandemic protocol modifications

shyamnaru:
| Q                   | A
| Bug fix?            | ❌ 
| New feature?        | ✅ 
| Database migration? | ❌ 
| New configuration?  | ❌ 
| Composer updates?   | ❌ 
| NPM updates?        | ❌ 
| Jira Ticket(s)      | HPRO-772,  HPRO-771

### Screenshots 
Screen Shot 2021-03-24 at 10 32 29 AM
Screen Shot 2021-03-24 at 10 32 50 AM
Screen Shot 2021-03-24 at 10 33 19 AM

### DIFF
--- a/src/Pmi/Evaluation/Evaluation.php
+++ b/src/Pmi/Evaluation/Evaluation.php
@@ -331,9 +331,6 @@ class Evaluation
 
             if ($type === 'radio') {
                 $options['expanded'] = true;
-                if (empty($this->data->{$field->name})) {
-                    $options['data'] = 'in-person';
-                }
             }
 
             if (isset($field->replicates)) {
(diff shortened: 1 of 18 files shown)


src/Pmi/Evaluation/Evaluation.php
```diff
@@ -272,6 +329,13 @@ public function getForm(FormFactory $formFactory)
             $options['constraints'] = $constraints;
             $options['attr'] = $attributes;
 
+            if ($type === 'radio') {
+                $options['expanded'] = true;
+                if (empty($this->data->{$field->name})) {
+                    $options['data'] = 'in-person';
+                }
```
jt2k, line 336: We currently don't have any other radio button fields, so this works, but can we find a way to make this work without specifying this specific value as the default here?

A few options that come to mind:
1. Have the default option value be the empty value. At least in our case here, we really only care about whether it's EHR or not. So a default empty value might work and be a more generalization solution. You might need to play around with the [`empty_data`](https://symfony.com/doc/current/reference/forms/types/choice.html#empty-data) option to get this to work.
2. Or, if it's better to have the in-person value be the actual option value, then maybe you can add an extra field in the JSON configuration to specify what the `empty_data` value is.

jt2k (approved):
Looks good! Thanks for getting this done so quickly! I haven't had a chance to pull this down and try it out yet, but I think this is good to merge into the feature branch and we can push to beta whenever you're ready.

stephenyeargin (approved):

PR #859, opened 2021-04-13
[HPRO-793] Fix Symfony timezone issues

shyamnaru:
| Q                   | A
| Target Release?     | 2.4.6.3 
| Bug fix?            | ✅
| New feature?        | ❌ 
| Database migration? | ❌ 
| New configuration?  | ❌ 
| Composer updates?   | ❌ 
| NPM updates?        | ❌ 
| Jira Ticket(s)      | HPRO-793

### DIFF
--- a/src/Pmi/Controller/DefaultController.php
+++ b/src/Pmi/Controller/DefaultController.php
@@ -39,7 +39,7 @@ class DefaultController extends AbstractController
 
     public function homeAction(Application $app)
     {
-        $checkTimeZone = $app->hasRole('ROLE_USER') || $app->hasRole('ROLE_ADMIN') || $app->hasRole('ROLE_AWARDEE') || $app->hasRole('ROLE_DV_ADMIN');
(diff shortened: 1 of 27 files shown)


stephenyeargin (commented):
I feel like there was a scenario where `User::$timezone` wasn't being set, and that's why we relied on the `info` block that was coming back. If we managed to mitigate that, this is definitely cleaner/more typical.

stephenyeargin:
`app.user.timezone` is also a little more concise than `app.user.getTimezone` in the Twig files. Twig automatically looks for the method with the `get` prefix.

shyamnaru:
> I feel like there was a scenario where `User::$timezone` wasn't being set, and that's why we relied on the `info` block that was coming back. If we managed to mitigate that, this is definitely cleaner/more typical.

I don't remember the reason why we used getInfo['timezone'] but getTimezone() seems to be working.

jt2k (approved):

stephenyeargin (approved):

PR #861, opened 2021-04-20
[HPRO-800] Update `@symfony/webpack-encore`

stephenyeargin:
| Q                   | A
| Bug fix?            | ✅ 
| New feature?        | ❌ 
| Database migration? | ❌ 
| New configuration?  | ❌ 
| Composer updates?   | ❌ 
| NPM updates?        | ✅ 
| Jira Ticket(s)      | HPRO-800 

### Summary

While only used in the asset build process, it is best to keep this up to date.


package-lock.json
```diff
@@ -1,8 +1,11038 @@
 {
     "name": "pmi-drc-hpo",
     "version": "1.0.0",
-    "lockfileVersion": 1,
+    "lockfileVersion": 2,
```
stephenyeargin, line 4: May need to ensure we're all up to npm v7.x for this to not flip back and forth.

jt2k (approved):

shyamnaru (approved):

jt2k:
Just un-approved this. Looked at this again, and it looks like the reason is-svg can't be updated automatically is because we're on an old version of webpack-encore. Can we update that instead of adding is-svg as a primary dependency?

jt2k:
Because we were using the pre 1.0 release, `npm update` isn't updating webpack encore. But if we just update to the latest (something like `npm install --dev "@symfony/webpack-encore@^1.1.2"`) I think that should take care of this.

stephenyeargin:
Verified that the build script still runs without issue.

```
$ npm run build

> pmi-drc-hpo@1.0.0 build /Users/yeargism/Sites/all-of-us-healthpro
> encore production

Running webpack ...

 DONE  Compiled successfully in 10861ms                                                                                                               2:29:50 PM

 I  46 files written to web/build
webpack compiled successfully
```

jt2k (approved):

PR #866, opened 2021-04-23
[HPRO-802] Format imported PatientStatus comments

stephenyeargin:
| Q                   | A
| Bug fix?            | ✅ 
| New feature?        | ❌ 
| Database migration? | ❌ 
| New configuration?  | ❌ 
| Composer updates?   | ❌ 
| NPM updates?        | ❌ 
| Jira Ticket(s)      | HPRO-802

### DIFF
--- a/symfony/src/Service/PatientStatusImportService.php
+++ b/symfony/src/Service/PatientStatusImportService.php
@@ -62,7 +62,7 @@ class PatientStatusImportService
             $row = [];
             $row['participantId'] = $patientStatusImportRow->getParticipantId();
             $row['patientStatus'] = $patientStatusImportRow->getStatus();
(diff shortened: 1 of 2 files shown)


jt2k (commented):
What do you think about using [this approach](https://datatables.net/manual/data/renderers#Text-helper) instead of manually escaping on the server side?

Outside the scope of this particular fix, we may want to look at applying the text renderer by default, and make an exception when we are specifically returning HTML.

stephenyeargin:
Had considered that as well, but looked to how we were handling it Work Queue as a guide.

I'm not sure there's a glob setting in DataTables we can enable. May result in needing to make a lot of other code changes to ensure it doesn't break anything we've been doing server side previously in other sections of the platform.

jt2k (approved):

jt2k (approved):
Nice!

shyamnaru (approved):

PR #654, opened 2020-05-07
[HPRO-599] Install symfony 4 framework

shyamnaru:

PR #868, opened 2021-04-30
[HPRO-812] Adjust cookie settings for SameSite attribute

stephenyeargin:
| Q                   | A
| Bug fix?            | ✅ 
| New feature?        | ❌ 
| Database migration? | ❌ 
| New configuration?  | ❌ 
| Composer updates?   | ❌ 
| NPM updates?        | ❌ 
| Jira Ticket(s)      | HPRO-812, PD-6062 

### Summary

Back in #654, we added the Symfony framework with the `SameSite` session cookie attribute to `lax`. We can adjust this to `strict`.

Reference: https://symfony.com/blog/new-in-symfony-4-2-samesite-cookie-configuration

### DIFF
--- a/symfony/config/packages/dev/framework.yaml
+++ b/symfony/config/packages/dev/framework.yaml
@@ -8,7 +8,7 @@ framework:
   session:
     handler_id: null
     cookie_secure: auto
-    cookie_samesite: lax
+    cookie_samesite: 'strict'
 
   #esi: true
   #fragments: true
--- a/symfony/config/packages/prod/framework.yaml
+++ b/symfony/config/packages/prod/framework.yaml
@@ -8,7 +8,7 @@ framework:
   session:
     handler_id: Pmi\Datastore\DatastoreSessionHandler
     cookie_secure: auto
-    cookie_samesite: lax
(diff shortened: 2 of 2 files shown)


jt2k:
Wouldn't we need to set this on the Silex side, too? At first, I was looking where we would configure that, which appears to be setting `$this['session.storage.options']` in `AbstractApplication::setup` right after we register the session provider (`            $this->register(new SessionServiceProvider());`)

But, then after digging more into it, and realizing that we aren't setting any cookie settings there, I realized we also have things set at the `php.ini` level:  
`session.cookie_httponly` is set to 1 and `session.cookie_secure` is set to 1 unless running locally.

Should we be setting the `samesite` attribute in `php.ini.dist` instead?

stephenyeargin:
> Should we be setting the `samesite` attribute in `php.ini.dist` instead?

I would imagine the application defaults would supersede what's in `php.ini`, right? Maybe set it in both places to ensure it doesn't get flipped?

jt2k (approved):

shyamnaru (approved):

PR #872, opened 2021-05-11
[HPRO-755] Migrate physical measurements to symfony

shyamnaru:
| Q                   | A
| Bug fix?            | ❌ 
| New feature?        | ❌ 
| Database migration? | ❌ 
| New configuration?  | ❌ 
| Composer updates?   | ❌ 
| NPM updates?        | ❌ 
| Jira Ticket(s)      | HPRO-755


symfony/src/Entity/MeasurementHistory.php
```diff
@@ -24,9 +24,9 @@ class MeasurementHistory
     private $evaluation;
 
     /**
-     * @ORM\Column(type="integer")
+     * @ORM\OneToOne(targetEntity="App\Entity\User", cascade={"persist", "remove"})
```
jt2k, line 27: Shouldn't this be a ManyToOne relationship since a user can create many measurement histories?

jt2k (commented):
This looks good! I had a question about one of the relationships, and then noticed that other existing relationships are similarly marked as OneToOne when I think they shouldn't be. Can you double check the relationships and make sure we have them defined properly?

symfony/src/Repository/MeasurementRepository.php
```diff
@@ -100,4 +100,24 @@ public function getUnloggedMissingMeasurements(): array
             'type' => MissingNotificationLog::MEASUREMENT_TYPE
         ]);
     }
+
+    public function getMeasurement($measurementId, $participantId)
+    {
+        $parentIds = $this->createQueryBuilder('m')
```
jt2k, line 106: When there are no PMs with parent ids, this returns an empty array, and the not in clause below results in the measurement not being found. I know this is the way it currently is pre-migration, but even if we fixed this to handle the empty result set, it's not ideal that we're having to load every modified PM id into an array to exclude them.

Can you remind me what the reason is for not returning a PM that has been superceded by another? Do we not allow the user to view these?

shyamnaru: We did this to mirror what we are doing for orders modifications. Since we are creating a new PM instead of modifying the existing one we decided to hide the parent PM.

shyamnaru: I went ahead and fixed the empty parent ids issue for now. We can revisit and see what would be a better solution for this.

jt2k (approved):

src/Pmi/Controller/SymfonyMigrationController.php
```diff
@@ -121,6 +125,24 @@ public function orderAction($participantId, $orderId, Application $app)
...
     }
 
+    /**
+     * @deprecated 2021-05-10
+     */
+    public function evaluationAction($participantId, $evalId, Application $app)
+    {
+        if (!$evalId) {
+            return $app->redirect(sprintf(
+                '/s/participant/%s/measurements',
```
stephenyeargin, line 135: Will we need trailing slash(es)?

stephenyeargin (approved):
Great stuff! Made one note on the legacy redirect, but otherwise looks 👍

src/Pmi/Controller/SymfonyMigrationController.php
shyamnaru: I don't think so since it's not an index route `/` like workqueue

PR #896, opened 2021-06-22
[HPRO-831] Assorted fixes from PHPStan

stephenyeargin:
I split this out into its own PR from #894 for discussion. I would need some guidance on how to tackle the three remaining items, and to ensure I didn't miss anything in my other fixes.


src/Pmi/Service/EmailNotificationService.php
```diff
@@ -1,135 +0,0 @@
-<?php
```
stephenyeargin, line 1: This appeared to no longer be in use, as everything related to it has been moved over to the Symfony namespace.

symfony/src/Controller/MeasurementsController.php
```diff
@@ -473,7 +473,7 @@ public function measurementFhirAction($participantId, $measurementId, Request $r
         }
         $fhir = $measurement->getFhir($date, $parentRdrId);
         if ($isTest) {
-            $fhir = \Tests\Pmi\Evaluation\EvaluationTest::getNormalizedFhir($fhir);
+            $fhir = \App\Test\Entity\MeasurementTest::getNormalizedFhir($fhir);
```
stephenyeargin, line 476: Note that this still appears in the `phpstan.baseline` file as still un-callable.

symfony/src/Service/EmailNotificationService.php
```diff
@@ -27,6 +27,12 @@ class EmailNotificationService
     protected $filterSummaries = false;
     protected $logEntity;
 
+    abstract protected function getSearchParams($id, $lastDeactivate);
+
+    protected function filterSummaries($summaries) {
```
stephenyeargin, line 32: This appears to be unreachable code anyway, but there wasn't a method defined for what should happen. Correct response may be to remove the logic entirely.

src/Pmi/Evaluation/Evaluation.php
```diff
@@ -283,7 +283,7 @@ public function getForm(FormFactory $formFactory)
                 $attributes['data-parsley-lt'] = $field->max;
             }
             if (isset($field->min)) {
-                $constraints[] = new Constraints\GreaterThanEqual($field->min);
+                $constraints[] = new Constraints\GreaterThanOrEqual($field->min);
```
stephenyeargin, line 286: This may have been a previous name of this class? Either way, I believe this was the intention.

src/Pmi/Cache/DatastoreAdapter.php
```diff
@@ -52,6 +52,7 @@ protected function doDelete(array $ids)
 
     protected function doSave(array $values, $lifetime)
     {
+        $failed = []; // Overwritten in method below (passed by reference)
```
stephenyeargin, line 55: Not sure how the Symfony library itself gets away with this one, but this made sense to me to squelch the item.

https://github.com/symfony/symfony/blob/fbd8d8ecb739724747ab31828e699e30d748de51/src/Symfony/Component/Cache/Adapter/CouchbaseBucketAdapter.php#L224-L242

shyamnaru (approved):

src/Pmi/Evaluation/Evaluation.php
jt2k: Looks like we've never actually had a field-level absolute minimum defined, so this has always been a bug. In the Symfony-migrated version, this is already fixed.

jt2k (approved):

PR #897, opened 2021-06-22
[HPRO-519] Add role selector to navbar

jt2k:
| Q                   | A
| Bug fix?            | ❌ 
| New feature?        | ✅ 
| Database migration? | ❌ 
| New configuration?  | ❌ 
| Composer updates?   | ❌ 
| NPM updates?        | ❌ 
| Jira Ticket(s)      | HPRO-519 

### Summary

Previously, our navbar has links to specific pages for regular site and awardee users, and then links to home pages for other roles (like admin, dashboard, biobank, etc.)

This PR replaces the main HealthPro home page link with a dropdown when a user has access to more than one role. Note that this frees up space in the navbar so in the future, we could add navbar links for different roles (like sites, page notices, etc. when in the admin context) but that is outside the scope of this PR.

I decided to implement this completely in the view layer since all of this is only affects the rendered navbar. This results in a bit more complexity in the view logic, but I think it's the right organization. Also, since we are still using the Silex homepage, I had to duplicate this across both templates. Looking forward to removing all of the Silex templates soon!

Finally, one unrelated change I made while testing switching roles - there was an error being thrown in dev (strict) mode for quanum orders when the address is empty. I fixed this by adding an `is defined` check.

### Instructions for testing

Try all combination of roles with different `gaBypassGroups` entries.

### Screenshots

Screen Shot 2021-06-22 at 4 16 52 PM

Screen Shot 2021-06-22 at 4 17 19 PM

### DIFF
--- a/symfony/templates/admin/base.html.twig
+++ b/symfony/templates/admin/base.html.twig
@@ -1,0 +1,2 @@
+{% set currentRoleNav = 'admin' %}
+{% extends 'base.html.twig' %}
--- a/symfony/templates/admin/debug/missing-measurements.html.twig
+++ b/symfony/templates/admin/debug/missing-measurements.html.twig
@@ -1,4 +1,4 @@
-{% extends 'base.html.twig' %}
+{% extends 'admin/base.html.twig' %}
 {% block title %}Physical Measurements - Missing RDR Data - {% endblock %}
 {% block body %}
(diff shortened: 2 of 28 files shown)


stephenyeargin (approved):
😻

shyamnaru (approved):

PR #908, opened 2021-07-14
[HPRO-764] Handle google auth failures

shyamnaru:
| Q                   | A
| Target Release?     | 2.5.4 
| Bug fix?            | ❌ 
| New feature?        | ❌ 
| Database migration? | ❌ 
| New configuration?  | ❌ 
| Composer updates?   | ❌ 
| NPM updates?        | ❌ 
| Jira Ticket(s)      | HPRO-764


symfony/src/Security/GoogleGroupsAuthenticator.php
```diff
@@ -69,18 +76,29 @@ public function checkCredentials($credentials, UserInterface $user)
             return true; // Bypass groups auth
         }
         $valid2fa = !($this->params->has('enforce2fa') && $this->params->get('enforce2fa')) || $user->hasTwoFactorAuth();
+        $this->session->set('authEmail', $user->getUsername());
+        if (!$valid2fa) {
+            $this->session->set('authFailureReason', '2fa');
+        } elseif (empty($user->getGroups())) {
+            $this->session->set('authFailureReason', 'groups');
+        }
```
jt2k, line 84: When we discussed earlier, I was thinking that you were going to create a new route to redirect to, which is why I suggested storing the reason in the session. Since you are rendering the template and returning the response directly in the failures method, these could just be stored in properties in this class, removing the need to write and read from the session.

symfony/templates/error-gae-auth.html.twig
```diff
@@ -0,0 +1,8 @@
+{% extends 'base.html.twig' %}
+{% block title %}Error - {% endblock %}
+{% block body %}
+    <div class="alert alert-danger">
+        <i class="fa fa-minus-circle"></i>
+        <strong>Access denied!</strong> <a href="{{ logoutUrl }}">Please try again</a>.
```
jt2k, line 6: Sorry, one more minor thing. Can we change this message to the one we previously used in the flash message ("Authentication failed. Please try again.") because if the OAuth process failed for some reason, they would see this message.

And can we change this template name to something else, since "gae" no longer applies? (A very long time ago, we were using an App Engine-specific auth process, instead of the regular Google OAuth workflow that we use now). Maybe name this one `error-auth.html.twig`, and then rename the other two to `error-auth-2fa.html.twig` and `error-auth-groups.html.twig`?

jt2k (approved):

PR #914, opened 2021-07-19
[HPRO-834] Build Google Group List and Navigation for Managed Access Dashboard

shyamnaru:
| Q                   | A
| Bug fix?            | ❌ 
| New feature?        | ✅
| Database migration? | ❌ 
| New configuration?  | ❌ 
| Composer updates?   | ❌ 
| NPM updates?        | ❌ 
| Jira Ticket(s)      | HPRO-834


symfony/templates/base.html.twig
```diff
@@ -126,6 +126,9 @@
                     {% if is_granted('ROLE_USER') and app.session.get('site') and app.session.get('orderType') == 'dv' and reportKitUrl %}
                         <li><a data-href="{{ reportKitUrl }}" class="external-link">Report Kit Problem</a></li>
                     {% endif %}
+                    {% if is_granted('ROLE_MANAGE_USERS') %}
+                        <li><a href="{{ path('access_manage_dashboard') }}">Manged Access Dashboard</a></li>
```
jt2k, line 130:
```suggestion
                        <li><a href="{{ path('access_manage_dashboard') }}">Managed Access Dashboard</a></li>
```

symfony/src/Service/GoogleGroupsService.php
```diff
@@ -126,4 +126,18 @@ public function getRole(string $userEmail, string $groupEmail): ?string
...
+        try {
+            $result = $this->callApi('members', 'listMembers', [$groupEmail]);
+            if ($result) {
+                return $result->getMembers();
+            } else {
+                return null;
+            }
+        } catch (GoogleException $e) {
+            return null;
+        }
```
jt2k, line 141: This try catch pattern is used in the getRole method above because we needed to handle the situation where the user is not in the group. In this case, we might need to think more about how we want to handle this if an error is returned. We might want to just not catch the exception at all, since anything other than a successful return of members is going to be an unexpected error.

At the very least, we should return an empty array instead of null if we want to silently fail.

symfony/templates/accessmanagement/group-members.html.twig
```diff
@@ -0,0 +1,45 @@
...
+        <li><a href="{{ path('access_manage_dashboard') }}">Managed Access Dashboard</a></li>
+        <li><a href="{{ path('access_manage_user_groups') }}">My Groups</a></li>
+        <li class="active">{{ group.name }}</li>
+    </ol>
+    <div class="page-header">
+        <h2><i class="fa fa-users" aria-hidden="true"></i> User Management</h2>
+    </div>
+    <h3>{{ group.name }}</h3>
+    <br/>
+    {% if app.session.get('managegroups') %}
```
jt2k, line 14: Why do we need this check? Isn't there already a role check for this route?

symfony/templates/accessmanagement/groups.html.twig
```diff
@@ -0,0 +1,44 @@
...
+        <li><a href="{{ path('symfony_home') }}">Home</a></li>
+        <li><a href="{{ path('access_manage_dashboard') }}">Managed Access Dashboard</a></li>
+        <li class="active">My Groups</li>
+    </ol>
+    <div class="page-header">
+        <h2><i class="fa fa-users" aria-hidden="true"></i> User Management</h2>
+    </div>
+    <br/>
+
+    {% if app.session.get('managegroups') %}
```
jt2k, line 13: Same as above - can we remove this?

jt2k (changes requested):

jt2k (approved):

stephenyeargin (approved):

PR #918, opened 2021-07-26
[HPRO-856] Google Groups member management

shyamnaru:
| Q                   | A
| Bug fix?            | ❌ 
| New feature?        | ✅ 
| Database migration? | ❌ 
| New configuration?  | ✅ 
| Composer updates?   | ❌ 
| NPM updates?        | ❌ 
| Jira Ticket(s)      | HPRO-856


dev_config/config.yml.dist
```diff
@@ -131,3 +131,6 @@ ehr_protocol_banner_message:
 
 # Enables physical measurements blood donor protocol modification workflow for diversion pouch sites
 # feature.blooddonorpm: false
+
+# When set only these sites will be able to access managed access dashbaord
+# feature.managegrouppilotsites:
```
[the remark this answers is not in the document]
shyamnaru: Handled in https://github.com/all-of-us/healthpro/pull/919 PR

symfony/src/Controller/AccessManagementController.php
```diff
@@ -2,17 +2,32 @@
...
+    const MEMBER_DOMAIN = '@pmi-ops.org';
+
+    private $googleGroupsService;
+    private $loggerService;
+
+    public function __construct(GoogleGroupsService $googleGroupsService, LoggerService $loggerService)
+    {
+        $this->googleGroupsService = $googleGroupsService;
+        $this->loggerService = $loggerService;
+    }
```
jt2k, line 29: 👍

symfony/src/Controller/AccessManagementController.php
```diff
@@ -32,16 +47,102 @@ public function userGroups()
...
+                            'member' => $member->email,
+                            'result' => 'success'
+                        ]);
+                        return $this->redirectToRoute('access_manage_user_group', ['groupId' => $groupId]);
+                    }
+                    $this->addFlash('error', $result['message']);
+                    $this->loggerService->log(Log::GROUP_MEMBER_REMOVE, [
+                        'member' => $member->email,
+                        'result' => 'fail'
+                    ]);
```
jt2k, line 133: I'm not sure we want to return the error message to the user. In the `removeMember` method, you are catching an API exception and returning the message. That may contain information that we don't want to display and/or could be confusing. I think a generic messsage could be added to the flash message, and then we can add the result message to the log object.

symfony/src/Controller/AccessManagementController.php
```diff
@@ -32,16 +47,102 @@ public function userGroups()
...
+                        'member' => $email,
+                        'result' => 'success'
+                    ]);
+                    return $this->redirectToRoute('access_manage_user_group', ['groupId' => $groupId]);
+                }
+                $this->addFlash('error', $result['message']);
+                $this->loggerService->log(Log::GROUP_MEMBER_ADD, [
+                    'member' => $email,
+                    'result' => 'fail'
+                ]);
```
jt2k, line 90: Same as below. Can we return a generic error message to the user, and then add the return the result message to the log object?

jt2k (changes requested):
Looks good overall! Just a couple notes on handling API call errors.

jt2k (approved):

PR #919, opened 2021-07-29
[HPRO-862] Email Notification for Users Removed from the Google Groups

shyamnaru:
| Q                   | A
| Bug fix?            | ❌ 
| New feature?        | ✅
| Database migration? | ❌ 
| New configuration?  | ❌ 
| Composer updates?   | ❌ 
| NPM updates?        | ❌ 
| Jira Ticket(s)      | HPRO-882

### DIFF
--- a/symfony/src/Service/AccessManagementService.php
+++ b/symfony/src/Service/AccessManagementService.php
@@ -43,6 +43,7 @@ class AccessManagementService
                 ])
                 ->send();
             $this->loggerService->log(Log::GROUP_MEMBER_REMOVE_NOTIFY, [
+                'group' => $group,
                 'member' => $member,
                 'memberLastDay' => $memberLastDay
             ]);
--- a/symfony/src/Service/AccessManagementService.php
+++ b/symfony/src/Service/AccessManagementService.php
@@ -1,0 +1,50 @@
+<?php
+
+namespace App\Service;
+
+use Symfony\Component\DependencyInjection\ParameterBag\ParameterBagInterface;
+use Twig\Environment;
+use Pmi\Audit\Log;
+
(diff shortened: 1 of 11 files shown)


web/assets/js/views/GroupMemberRemove.js
```diff
@@ -0,0 +1,38 @@
...
+
+    var removeReason = 'input[name="remove_group_member[reason]"]';
+    showHideMemberLastDayField(removeReason);
+    $(removeReason).on('change', function () {
+        showHideMemberLastDayField(removeReason);
+    });
+
+    $('#remove_group_member_memberLastDay').pmiDateTimePicker({
+        format: 'MM/DD/YYYY',
+        maxDate: Date(),
```
stephenyeargin, line 35: It looks like our library will accept either, but it's more expected to use the class version of this `new Date()` which returns an object rather than `Date()` which returns a string.

symfony/src/Service/AccessManagementService.php
```diff
@@ -0,0 +1,51 @@
...
+                    'group' => $group,
+                    'member' => $member,
+                    'memberLastDay' => $memberLastDay->format('m/d/Y'),
+                    'loggedUser' => $this->userService->getUser()->getEmail(),
+                    'currentTime' => $currentTime->format('Y-m-d H:i:s e')
+                ])
+                ->send();
+            $this->loggerService->log(Log::GROUP_MEMBER_REMOVE_NOTIFY, [
+                'member' => $member,
+                'memberLastDay' => $memberLastDay
```
jt2k, line 47: Can we add the group to the log data, too?

stephenyeargin (approved):

PR #928, opened 2021-08-06
[HPRO-877] Allow blood donor PM collection for configured sites

shyamnaru:
| Q                   | A
| Target Release?     | 2.5.5 
| Bug fix?            | ❌ 
| New feature?        | ✅ 
| Database migration? | ❌ 
| New configuration?  | ✅ 
| Composer updates?   | ❌ 
| NPM updates?        | ❌ 
| Jira Ticket(s)      | HPRO-877


symfony/src/Service/MeasurementService.php
```diff
@@ -96,7 +96,16 @@ public function getMeasurmeent($participantId, $measurementId)
...
+        return $this->params->has('feature.blooddonorpm') && $this->params->get('feature.blooddonorpm') && $this->session->get('siteType') === 'dv' && ($this->siteService->isDiversionPouchSite() || $this->isBloodDonorPmSite());
+    }
+
+    public function isBloodDonorPmSite(): bool
+    {
+        if ($this->params->has('feature.blooddonorpmsites') && $sites = $this->params->get('feature.blooddonorpmsites')) {
+            $sites = explode(',', $sites);
+            return in_array($this->siteService->getSiteId(), $sites);
+        }
+        return false;
```
jt2k, line 108: Did you consider putting this method in the site service class? The diversion pouch site check seems similar, and might make sense to organize together.

symfony/src/Service/MeasurementService.php
```diff
@@ -96,7 +96,16 @@ public function getMeasurmeent($participantId, $measurementId)
 
     public function requireBloodDonorCheck()
     {
-        return $this->params->has('feature.blooddonorpm') && $this->params->get('feature.blooddonorpm') && $this->session->get('siteType') === 'dv' && $this->siteService->isDiversionPouchSite();
+        return $this->params->has('feature.blooddonorpm') && $this->params->get('feature.blooddonorpm') && $this->session->get('siteType') === 'dv' && ($this->siteService->isDiversionPouchSite() || $this->isBloodDonorPmSite());
```
jt2k, line 99: I think we discussed this previously, but just wanted to double check - did we decide that if a site is not in a DV awardee that it shouldn't have the abbreviated PM form, even if the site id is included in the `blooddonorpmsites` list?

shyamnaru: Yes, I remember we decided to enable this only when the site's type is DV because the requirement is to have blood donor pm form and dv kit order workflow (which only applies when the site's type is DV)

jt2k (approved):

PR #935, opened 2021-08-19
[HPRO-856] Managed access dashboard new changes

shyamnaru:
| Q                   | A
| Target Release?     | 2.6.0 
| Bug fix?            | ❌ 
| New feature?        | ✅
| Database migration? | ❌ 
| New configuration?  | ❌ 
| Composer updates?   | ❌ 
| NPM updates?        | ❌ 
| Jira Ticket(s)      | HPRO-856

### DIFF
--- a/symfony/src/Controller/AccessManagementController.php
+++ b/symfony/src/Controller/AccessManagementController.php
@@ -75,21 +75,25 @@ class AccessManagementController extends AbstractController
         if ($groupMemberForm->isSubmitted()) {
             if ($groupMemberForm->isValid()) {
                 $email = $groupMemberForm->get('email')->getData() . self::MEMBER_DOMAIN;
-                $result = $this->googleGroupsService->addMember($group->email, $email);
-                if ($result['status'] === 'success') {
-                    $this->addFlash('success', $result['message']);
-                    $this->loggerService->log(Log::GROUP_MEMBER_ADD, [
-                        'member' => $email,
-                        'result' => 'success'
-                    ]);
-                    return $this->redirectToRoute('access_manage_user_group', ['groupId' => $groupId]);
+                if ($this->googleGroupsService->getUser($email)) {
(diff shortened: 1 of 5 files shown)


symfony/src/Controller/AccessManagementController.php
```diff
@@ -75,21 +77,25 @@ public function member($groupId, Request $request)
...
+                        $this->addFlash('success', $result['message']);
+                        $this->loggerService->log(Log::GROUP_MEMBER_ADD, [
+                            'member' => $email,
+                            'result' => 'success'
+                        ]);
+                        return $this->redirectToRoute('access_manage_user_group', ['groupId' => $groupId]);
+                    }
+                    $errorMessage = isset($result['code']) && $result['code'] === 409 ? 'Member already exists.' : 'Error occurred. Please try again.';
+                } else {
+                    $errorMessage = 'User not registred.';
```
jt2k, line 92: Typo here. Can we also reword this to something like "User does not exist?"

jt2k, line 92: Can we move this outside the isValid block, and then handle these as form errors tied to the username field?

symfony/src/Service/GoogleGroupsService.php
```diff
@@ -141,6 +141,19 @@ public function getMembers(string $groupEmail)
...
+    public function getUser(string $user)
+    {
+        try {
+            $result = $this->callApi('users', 'get', [$user]);
+            if ($result) {
+                return $result;
+            }
+            return [];
+        } catch (GoogleException $e) {
+            return [];
```
jt2k, line 153: Would it make more sense to return null instead of an empty array, since this is getting a single user object, if it exists?

jt2k (changes requested):

jt2k (approved):

PR #946, opened 2021-09-01
[HPRO-906] Preserve previously set date

stephenyeargin:
| Q                   | A
| Bug fix?            | ✅ 
| New feature?        | ❌ 
| Database migration? | ❌ 
| New configuration?  | ❌ 
| Composer updates?   | ❌ 
| NPM updates?        | ❌ 
| Jira Ticket(s)      | HPRO-906 

### Summary

Only run `.pmiDateTimePicker()` against non-finalized orders.

### Screenshots

Screen Shot 2021-09-03 at 6 59 01 PM

### DIFF
--- a/web/assets/js/views/PhysicalEvaluation-0.3-ehr.js
+++ b/web/assets/js/views/PhysicalEvaluation-0.3-ehr.js
@@ -738,12 +738,14 @@ PMI.views['PhysicalEvaluation-0.3-ehr'] = Backbone.View.extend({
         this.render();
         var today = new Date();
         var sixMonthsAgo = (new Date()).setMonth(today.getMonth() - 6);
-        $('input.ehr-date').not(':disabled').pmiDateTimePicker({
-            'format': 'MM/DD/YYYY',
-            'maxDate': today,
-            'minDate': sixMonthsAgo,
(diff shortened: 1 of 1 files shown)


web/assets/js/views/PhysicalEvaluation-0.3-ehr.js
```diff
@@ -738,7 +738,7 @@ PMI.views['PhysicalEvaluation-0.3-ehr'] = Backbone.View.extend({
         this.render();
         var today = new Date();
         var sixMonthsAgo = (new Date()).setMonth(today.getMonth() - 6);
-        $('.ehr-date').pmiDateTimePicker({
+        $('input.ehr-date').not(':disabled').pmiDateTimePicker({
```
jt2k, line 741: I don't know if this applies to this field, but there are situations where some fields on the form may be hidden or disabled due to branching logic. If the form was initially loaded with a date field disabled, and then other changes caused the field to be enabled, the date picker wouldn't be applied. Would it make more sense to check `if (!this.finalized)`?

shyamnaru (approved):

jt2k (approved):

PR #955, opened 2021-09-15
[HPRO-911] Fix selecting current date in EHR physical measurements and deceased report page

shyamnaru:
| Q                   | A
| Bug fix?            | ✅ 
| New feature?        | ❌ 
| Database migration? | ❌ 
| New configuration?  | ❌ 
| Composer updates?   | ❌ 
| NPM updates?        | ❌ 
| Jira Ticket(s)      | HPRO-911

### DIFF
--- a/symfony/src/Form/MeasurementType.php
+++ b/symfony/src/Form/MeasurementType.php
@@ -96,7 +96,7 @@ class MeasurementType extends AbstractType
                 $class = Type\DateType::class;
                 $minDate = new \DateTime('-6 months');
                 $constraints[] = new Constraints\Range([
-                    'min' => $minDate->modify('-1 day')->setTime(0,0,0),
+                    'min' => $minDate->modify('-1 day')->setTime(0, 0, 0),
                     'max' => new \DateTime('today'),
                     'notInRangeMessage' => 'Date cannot be greater than six months in the past and cannot be in the future'
                 ]);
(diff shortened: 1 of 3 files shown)


stephenyeargin (commented):
I had thought that the trouble that came up in testing was on the PHP side rather than the JavaScript side. For example, March 8 was allowed by the date picker library, but PHP saw it as `2021-03-08 00:00` was less than the minimum of `2021-03-08 09:17` (precisely six months prior).

image

stephenyeargin (approved):

symfony/src/Form/MeasurementType.php
```diff
@@ -94,11 +94,11 @@ public function buildForm(FormBuilderInterface $builder, array $options)
             } elseif ($type === 'date') {
                 unset($fieldOptions['scale']);
                 $class = Type\DateType::class;
+                $minDate = new \DateTime('-6 months');
```
jt2k, line 97: For clarity, can you modify $minDate here and then use in the constraint below, instead of modifying it in the same line where it is assigned?

symfony/src/Form/MeasurementType.php
```diff
@@ -94,11 +94,11 @@ public function buildForm(FormBuilderInterface $builder, array $options)
...
                 $constraints[] = new Constraints\Range([
-                    'min' => new \DateTime('-6 months'),
+                    'min' => $minDate->modify('-1 day')->setTime(0, 0, 0),
                     'max' => new \DateTime('today'),
-                    'minMessage' => 'Date cannot be greater than six months in the past',
-                    'maxMessage' => 'Date cannot be in the future'
```
jt2k, line 101: The min and max message are valid options for the range constraint. It's not clear to me why they aren't used instead of the notInRangeMessage, but can we keep them here? If it's potentially a bug in the validator component that is later fixed, we want our custom messages to still be displayed.

jt2k (changes requested):

jt2k (approved):

PR #960, opened 2021-09-23
[HPRO-917] Adjust routing rules for `/s/cron` endpoints

stephenyeargin:
| Q                   | A
| Bug fix?            | ✅ 
| New feature?        | ❌ 
| Database migration? | ❌ 
| New configuration?  | ❌ 
| Composer updates?   | ✅ 
| NPM updates?        | ❌ 
| Jira Ticket(s)      | HPRO-917 

### Summary

Adjusts routing rules around `/s/cron` endpoints.

### Testing

Provide the `X-Appengine-Cron: true` header when interacting with the endpoints. Omitting it will return a 404 error.

Example:

```
$ curl -H 'X-Appengine-Cron: true' http://localhost:8080/s/cron/ping-test
```

### DIFF
--- a/symfony/src/Controller/CronController.php
+++ b/symfony/src/Controller/CronController.php
@@ -20,7 +20,7 @@ use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\Routing\Annotation\Route;
 
 /**
- * @Route("/s/cron", condition="request.headers.get('X-Appengine-Cron') matches '/true/'")
+ * @Route("/s/cron", condition="request.headers.get('X-Appengine-Cron') === 'true'")
  */
 class CronController extends AbstractController
 {
--- a/symfony/src/Controller/CronController.php
+++ b/symfony/src/Controller/CronController.php
@@ -20,7 +20,7 @@ use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\Routing\Annotation\Route;
 
 /**
- * @Route("/s/cron")
+ * @Route("/s/cron", condition="request.headers.get('X-Appengine-Cron') matches '/true/'")
  */
 class CronController extends AbstractController
 {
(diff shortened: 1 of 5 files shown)


symfony/src/Controller/CronController.php
```diff
@@ -20,7 +20,7 @@
 use Symfony\Component\Routing\Annotation\Route;
 
 /**
- * @Route("/s/cron")
+ * @Route("/s/cron", condition="request.headers.get('X-Appengine-Cron') matches '/true/'")
```
jt2k, line 23: What's the reason for a regex match here? Do we not want the exact string 'true'?

shyamnaru:
Previously we are not running this check locally https://github.com/all-of-us/healthpro/blob/0729b0ea34964668c602d676d30f689b91e137bd/src/Pmi/Controller/CronController.php#L19-L22. This change will prevent hitting the cron urls from the local browser. As you mentioned we can use curl to run the cron jobs.

symfony/src/Controller/CronController.php
stephenyeargin: Example code only used `matches`, so I'd assumed that it had to be regular expression. Further reading indicates I can get away with an exact string. (`==` or `===`)

stephenyeargin:
>  This change will prevent hitting the cron urls from the local browser. As you mentioned we can use curl to run the cron jobs.

I'd considered this as well. Because it's particularly important that we make sure this operates behind the check, I didn't want to make an exception for local development. There are a handful of plugins to modify the request headers your browser sends to continue to be able to test calls that way.

Example: https://github.com/bewisse/modheader

symfony/src/Controller/CronController.php
stephenyeargin: Resolved in 6c125e4.

jt2k (approved):

shyamnaru (approved):

PR #966, opened 2021-10-01
[HPRO-895] Add debug logs to keepalive CSRF token mismatch

stephenyeargin:
| Q                   | A
| Bug fix?            | ✅ 
| New feature?        | ❌ 
| Database migration? | ❌ 
| New configuration?  | ❌ 
| Composer updates?   | ❌ 
| NPM updates?        | ❌ 
| Jira Ticket(s)      | HPRO-895 

### Summary

Looking to collect some more context around the HTTP 403 errors we're seeing for certain users where the CSRF token doesn't match the submitted value.

### DIFF
--- a/symfony/src/Controller/DefaultController.php
+++ b/symfony/src/Controller/DefaultController.php
@@ -80,7 +80,7 @@ class DefaultController extends AbstractController
      */
     public function keepAliveAction(Request $request, LoggerService $loggerService)
     {
-        if ($this->isCsrfTokenValid('keepAlive', $request->request->get('csrf_token'))) {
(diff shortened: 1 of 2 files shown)


stephenyeargin:
Is there anything else we might want to collect?

jt2k (approved):
Looks good to me. I did consider whether we should add the expected token value with something like:

```php
'expected_token' => $this->get('security.csrf.token_manager')->getToken('keepAlive')->getValue(),
```

but I'm not sure how much value that has, because that method will regenerate a new token if it doesn't already exist in storage.

PR #976, opened 2021-10-22
[HPRO-934] Implement WQ drop down navigation

shyamnaru:
| Q                   | A
| Bug fix?            | ❌ 
| New feature?        | ✅
| Database migration? | ❌ 
| New configuration?  | ❌ 
| Composer updates?   | ❌ 
| NPM updates?        | ❌ 
| Jira Ticket(s)      | HPRO-934


symfony/templates/workqueue/consents.html.twig
```diff
@@ -41,11 +41,14 @@
             {% include 'partials/workqueue-consents-advance-filters.html.twig' %}
         </div>
         <div class="col-md-9">
-            {% if canExport %}
-                <p class="text-right">
+            <div class="text-right">
+                <div class="page-drop-down">
+                    Page <select name="pageDropDown" class="input-sm"></select>&nbsp;of&nbsp;<span class="total-pages"></span>
```
jt2k, line 46: This select doesn't match the style of all our other dropdowns. I'm guessing you might have not added the `form-control` class because it makes the select a block element. But if you add the `form-inline` class to one of the parent divs, it works pretty well.

So, something like:

```suggestion
            <div class="text-right form-inline">
                <div class="page-drop-down">
                    Page <select name="pageDropDown" class="form-control input-sm"></select>&nbsp;of&nbsp;<span class="total-pages"></span>
```

And since this is repeated four times (twice on each work queue view), can you move this whole block to a separate partial template?

symfony/templates/workqueue/consents.html.twig
```diff
@@ -41,14 +41,7 @@
...
         <div class="col-md-9">
-            <div class="text-right form-inline">
-                <div class="page-drop-down">
-                    Page <select name="pageDropDown" class="form-control input-sm"></select>&nbsp;of&nbsp;<span class="total-pages"></span>
-                </div>
-                {% if canExport %}
-                    <button data-href="{{ path('workqueue_export', params) }}" class="btn btn-xs btn-default export"><i class="fa fa-download" aria-hidden="true"></i> Export</button>
-                {% endif %}
-            </div>
+            {% include 'partials/workqueue-drop-down-navigation-and-eport-button.html.twig' %}
```
jt2k, line 44: Typo here (eport). But can we also shorten the file name? Maybe something like `workqueue-page-export.html.twig`?

stephenyeargin (approved):

PR #989, opened 2021-11-15
[HPRO-890] Cleanup composer.json file and update dependencies

shyamnaru:
| Q                   | A
| Bug fix?            | ❌ 
| New feature?        | ❌ 
| Database migration? | ❌ 
| New configuration?  | ❌ 
| Composer updates?   | ✅/
| NPM updates?        | ❌ 
| Jira Ticket(s)      | HPRO-890


composer.lock
```diff
@@ -2423,68 +2974,59 @@
...
+            "support": {
+                "issues": "https://github.com/maxmind/web-service-common-php/issues",
+                "source": "https://github.com/maxmind/web-service-common-php/tree/v0.8.1"
+            },
+            "time": "2020-11-02T17:00:53+00:00"
         },
         {
             "name": "monolog/monolog",
-            "version": "2.3.4",
+            "version": "1.26.1",
```
jt2k, line 2985: The Symfony monolog bundle was downgraded from 5.x to 4.4, which is what we want. But a side effect of this is a major version downgrade for monolog. We should check to make sure this doesn't have any unintended side effects.

stephenyeargin (approved):

composer.lock
```diff
@@ -1483,11 +1901,14 @@
                 }
             },
             "autoload": {
-                "psr-0": {
-                    "Google_": "src/"
```
jt2k, line 1487: I think we have a number of places where we are still using the old PSR-0 `Google_`... class names. I guess the aliases file makes this backwards compatible? Can you confirm that all of our Google client functionality that is using the underscore separated global class names still works?

composer.json
```diff
@@ -50,11 +55,12 @@
...
         "phpunit/phpunit": "^9",
         "symfony/maker-bundle": "^1.15",
-        "symfony/profiler-pack": "^1.0",
         "phpstan/phpstan": "^0.12.89",
         "phpstan/phpstan-symfony": "^0.12.36",
         "phpstan/phpstan-doctrine": "^0.12.39",
-        "phpstan/extension-installer": "^1.1"
+        "phpstan/extension-installer": "^1.1",
+        "symfony/stopwatch": "^4.4",
+        "symfony/web-profiler-bundle": "^4.4"
```
jt2k, line 63: Same with these, too.

jt2k (changes requested):

jt2k (approved):

stephenyeargin (approved):

PR #1002, opened 2021-11-23
[HPRO-963] Adjust gRoR column to link to PDF consent

stephenyeargin:
| Q                   | A
| Bug fix?            | ✅ 
| New feature?        | ❌ 
| Database migration? | ❌ 
| New configuration?  | ❌ 
| Composer updates?   | ❌ 
| NPM updates?        | ❌ 
| Jira Ticket(s)      | HPRO-963 

### Summary

The Genetic Return of Results (gRoR) column should use the `displayConsentStatus` method for rendering so that it can pick up the PDF link if available.

### DIFF
--- a/src/Helper/WorkQueue.php
+++ b/src/Helper/WorkQueue.php
@@ -1007,22 +1007,6 @@ class WorkQueue
         return '';
     }
 
-    public static function displayGenomicsConsentStatus($value, $time, $userTimezone, $displayTime = true)
-    {
-        switch ($value) {
-            case 'SUBMITTED':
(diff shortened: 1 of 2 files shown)


jt2k (changes requested):
This fixes the consent work queue, but the main work queue is still using the `displayGenomicsConsentStatus` method which doesn't include the link.

Looks like the following additional changes are required:

* Update the `WorkQueueService::generateTableRows` to use `displayConsentStatus` instead of `displayGenomicsConsentStatus` for the `gRoRConsent` row.
* Remove the `WorkQueue::displayGenomicsConsentStatus` method which should no longer be used anywhere.

jt2k (approved):

shyamnaru (approved):

PR #1021, opened 2021-12-22
[HPRO-976] Display CABoR Consent for California sites only

shyamnaru:
| Q                   | A
| Bug fix?            | ❌ 
| New feature?        | ✅ 
| Database migration? | ❌ 
| New configuration?  | ❌ 
| Composer updates?   | ❌ 
| NPM updates?        | ❌ 
| Jira Ticket(s)      | HPRO-976


tests/Service/SiteServiceTest.php
```diff
@@ -0,0 +1,41 @@
...
+        $this->id = uniqid();
+        $site = 'hpo-site-test' . $this->id;
+        $this->login('test@example.com', [$site]);
+        $this->createSite();
+        $this->service = static::$container->get(SiteService::class);
+        $this->service->switchSite($site . '@' . self::GROUP_DOMAIN);
+        self::assertFalse($this->service->displayCaborConsent());
+    }
+
+    public function createSite(): void
```
jt2k, line 25: Since this is only used inside other public test methods, can you make this private?

tests/Service/ServiceTestCase.php
```diff
@@ -0,0 +1,52 @@
...
+use App\Service\MockGoogleGroupsService;
+use App\Service\UserService;
+use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase;
+use Symfony\Component\HttpFoundation\Session\SessionInterface;
+use Symfony\Component\Security\Core\Authentication\Token\PreAuthenticatedToken;
+use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface;
+use Symfony\Component\Security\Core\User\UserProviderInterface;
+use App\Tests\GoogleGroup;
+
+class ServiceTestCase extends KernelTestCase
```
jt2k, line 14: Should this be an abstract class?

src/Service/SiteSyncService.php
```diff
@@ -140,6 +140,9 @@ public function sync($preview = false)
                     $siteData->setTimezone(isset($site->timeZoneId) ? $site->timeZoneId : null);
                     $siteData->setType($awardee->type);
                     $siteData->setSiteType(isset($site->siteType) ? $site->siteType : null);
+                    if (isset($site->address->state)) {
```
jt2k, line 143: If there's the possibility of an address object without a state field, is there also the possibility of a site object without an address? That is, should we check for `(isset($site->address) && isset($site->address->state))`?

jt2k (changes requested):

src/Service/SiteSyncService.php
shyamnaru: Yes, I guess it doesn't hurt to check if address field is set.

jt2k (approved):

src/Service/SiteService.php
```diff
@@ -303,4 +307,9 @@ public function getSiteType()
     {
         return $this->isDVType() ? 'dv' : 'hpo';
     }
+
+    public function displayCaborConsent(): bool
+    {
+        return $this->session->get('siteState') === self::CABOR_STATE ? false : true;
```
jt2k, line 313: Isn't this backwards? CABoR consent should be displayed for CA sites, and only for CA sites.

tests/Service/SiteServiceTest.php
```diff
@@ -0,0 +1,41 @@
...
+use App\Entity\Site;
+use App\Service\SiteService;
+use Doctrine\ORM\EntityManagerInterface;
+
+class SiteServiceTest extends ServiceTestCase
+{
+    protected $service;
+    protected $id;
+
+    public function testCaborConsentDisplay(): void
```
jt2k, line 14: After inverting the logic (see comment above), let's also make sure we are testing both scenarios - a site in CA and a site not in CA.

jt2k (changes requested):

src/Service/SiteService.php
shyamnaru: Yes, you are right. I guess I got tricked by the session variable not being set while testing.

jt2k (approved):

PR #1026, opened 2022-01-05
[HPRO-975] Add PID & Name to Participant Review tabs

shyamnaru:
| Q                   | A
| Bug fix?            | ❌ 
| New feature?        | ✅
| Database migration? | ❌ 
| New configuration?  | ❌ 
| Composer updates?   | ❌ 
| NPM updates?        | ❌ 
| Jira Ticket(s)      | HPRO-975

### DIFF
--- a/src/Controller/ReviewController.php
+++ b/src/Controller/ReviewController.php
@@ -123,6 +123,7 @@ class ReviewController extends AbstractController
         $unlockedOrders = $orderRepository->getSiteUnlockedOrders($site);
         $unfinalizedOrders = $orderRepository->getSiteUnfinalizedOrders($site);
         $orders = array_merge($unlockedOrders, $unfinalizedOrders);
+        $this->reviewService->loadFirstFiveParticipantNames($orders);
         return $this->render('review/unfinalized_orders.html.twig', [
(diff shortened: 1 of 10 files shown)


src/Controller/ReviewController.php
```diff
@@ -123,6 +123,7 @@ public function unfinalizedOrders(Request $request, OrderRepository $orderReposi
         $unlockedOrders = $orderRepository->getSiteUnlockedOrders($site);
         $unfinalizedOrders = $orderRepository->getSiteUnfinalizedOrders($site);
         $orders = array_merge($unlockedOrders, $unfinalizedOrders);
+        $this->reviewService->loadFirstFiveParticipantNames($orders);
```
jt2k, line 126: Would it be possible to use this method in the `today` method, also?

jt2k (approved):
Looks good! Just one question about how easy it would be to reuse this new method in the today endpoint, also.

src/Controller/ReviewController.php
shyamnaru: Thought about it but the `getTodayParticipants` method returns a different array format I decided to use it only for other review tabs.

PR #1027, opened 2022-01-07
[HPRO-958] Adjust log level for 404 not found exceptions

shyamnaru:
| Q                   | A
| Bug fix?            | ❌ 
| New feature?        | ✅ 
| Database migration? | ❌ 
| New configuration?  | ❌ 
| Composer updates?   | ❌ 
| NPM updates?        | ❌ 
| Jira Ticket(s)      | HPRO-958

### DIFF
--- a/src/Service/LoggerService.php
+++ b/src/Service/LoggerService.php
@@ -7,6 +7,7 @@ use Psr\Log\LoggerInterface;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\RequestStack;
 use Symfony\Component\HttpFoundation\Session\SessionInterface;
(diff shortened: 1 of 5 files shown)


src/Service/LoggerService.php
```diff
@@ -18,47 +19,6 @@ class LoggerService
     protected $action;
     protected $data;
 
-    public const PMI_AUDIT_PREFIX = 'PMI_AUDIT_';
```
jt2k, line 21: This is pretty minor, but I think it might make sense to keep this constant here. The prefix is only needed internally by this class.

jt2k (approved):
Looks good! One suggestion about the prefix constant, but not required if you prefer keeping it in the Log class.

PR #1030, opened 2022-01-11
[HPRO-1000] Increase phpstan level to 1

shyamnaru:
| Q                   | A
| Bug fix?            | ❌ 
| New feature?        | ✅
| Database migration? | ❌ 
| New configuration?  | ❌ 
| Composer updates?   | ❌ 
| NPM updates?        | ❌ 
| Jira Ticket(s)      | HPRO-1000

### DIFF
--- a/src/Controller/ProblemController.php
+++ b/src/Controller/ProblemController.php
@@ -209,9 +209,9 @@ class ProblemController extends AbstractController
         }
         if ($problemId) {
             $problem = $problemRepository->findOneBy(['id' => $problemId, 'participantId' => $participantId]);
-            if (!$problem && !$problem->getFinalizedTs()) {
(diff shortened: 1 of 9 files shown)


src/Controller/ProblemController.php
```diff
@@ -209,9 +209,9 @@ public function problemComment($participantId, $problemId, Request $request, Log
         }
         if ($problemId) {
```
jt2k, line 210: Could we instead remove this check? The problemId has to be set, since it's part of the route, and doesn't have a default value. If you remove that if, then $problem will always be set to the result of the `findOneBy` and we can continue using the same check as before.

jt2k (approved):
🆙

PR #1056, opened 2022-02-07
[HPRO-1001] Add New Syncope Question for Biobank Workflow

shyamnaru:
| Q                   | A
| Bug fix?            | ❌ 
| New feature?        | ✅ 
| Database migration? | ❌ 
| New configuration?  | ❌ 
| Composer updates?   | ❌ 
| NPM updates?        | ❌ 
| Jira Ticket(s)      | HPRO-1001

### DIFF
--- a/templates/order/check.html.twig
+++ b/templates/order/check.html.twig
@@ -34,12 +34,14 @@
         </div>
 
         <div id="transfusion">
-            <h3>In the past 6 months, has the participant had a blood transfusion?</h3>
-            <div class="radio">
-                <label><input type="radio" name="transfusion" value="yes"/> Yes</label>
-            </div>
-            <div class="radio">
-                <label><input type="radio" name="transfusion" value="no"/> No</label>
+            <div>
+                <h3>In the past 6 months, has the participant had a blood transfusion?</h3>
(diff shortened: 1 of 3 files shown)


templates/order/check.html.twig
```diff
@@ -71,6 +73,24 @@
...
+                <div class="radio">
+                    <label><input type="radio" name="syncope" value="no"/> No</label>
+                </div>
+                <div style="display:none;" id="syncope-sub-qn">
+                    <h3>Would the participant like to continue with a full HPO Biobank order or a modified order with urine and saliva only?</h3>
+                    <div class="radio">
+                        <label><input type="radio" name="syncope_sub" value="yes"/> Full HPO Biobank Order</label>
+                    </div>
+                    <div class="radio">
+                        <label><input type="radio" name="syncope_sub" value="no"/> Modified Order - {% if app.session.get('siteType') != 'dv' %} Saliva and Urine {% endif %} Only</label>
```
jt2k, line 90: It looks like we were already previously doing this, but would it make sense to pass in the site type to the view from the controller, instead of accessing the session directly?

jt2k (approved):

PR #1058, opened 2022-02-08
[HPRO-1052] Remove settings, problems, and participant view JS files in twig

shyamnaru:
| Q                   | A
| Bug fix?            | ❌ 
| New feature?        | ✅/
| Database migration? | ❌ 
| New configuration?  | ❌ 
| Composer updates?   | ❌ 
| NPM updates?        | ❌ 
| Jira Ticket(s)      | HPRO-1052

### DIFF
--- a/templates/participant/details.html.twig
+++ b/templates/participant/details.html.twig
@@ -2,7 +2,8 @@
 {% block title %}Participant {{ participant.id }} - {% endblock %}
 {% block body %}
     {% import 'macros/display-text.html.twig' as macros %}
-    <div id="participant_info" data-participant-id="{{ participant.id }}" data-org-patient-status="{{ orgPatientStatusData|json_encode }}" data-can-view-patient-status="{{ canViewPatientStatus }}">
+    {% set hasOrgPatientStatusData = orgPatientStatusData ? 1 : 0 %}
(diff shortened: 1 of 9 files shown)


templates/participant/details.html.twig
```diff
@@ -2,6 +2,7 @@
 {% block title %}Participant {{ participant.id }} - {% endblock %}
 {% block body %}
     {% import 'macros/display-text.html.twig' as macros %}
+    <div id="participant_info" data-participant-id="{{ participant.id }}" data-org-patient-status="{{ orgPatientStatusData|json_encode }}" data-can-view-patient-status="{{ canViewPatientStatus }}">
```
jt2k, line 5: It looks like the `orgPatientStatusData` is only used to determine whether there are any patient status data or not. Instead of passing in the entire structure, could you just determine here whether or not it's empty and then set it to false/true or 0/1 or something like that?

jt2k: Alternatively, if you're going to keep the entire structure here, I think you might need to change the escape strategy to: `orgPatientStatusData|json_encode|escape('html_attr')` since it's inside an attribute.

jt2k (changes requested):

jt2k (approved):

PR #1060, opened 2022-02-14
[HPRO-947] Integrate google secret manager

shyamnaru:
| Q                   | A
| Bug fix?            | ❌ 
| New feature?        | ✅ 
| Database migration? | ❌ 
| New configuration?  | ❌ 
| Composer updates?   | ❌ 
| NPM updates?        | ❌ 
| Jira Ticket(s)      | HPRO-947

### DIFF
--- a/src/SecretManager/SecretManager.php
+++ b/src/SecretManager/SecretManager.php
@@ -6,6 +6,9 @@ use Google\Cloud\SecretManager\V1\SecretManagerServiceClient;
 
 class SecretManager
 {
+    private const SECRET_NAME_CREDENTIALS = 'credentials';
+    private const SECRET_NAME_GA_AUTH_JSON = 'gaAuthJson';
+
     private $secretManagerClient;
 
     public function __construct()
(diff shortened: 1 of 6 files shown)


src/SecretManager/SecretManager.php
```diff
@@ -0,0 +1,53 @@
...
+        if (!$useDefaultCredentials) {
+            $basePath = realpath(__DIR__ . '/../../');
+            $keyFile = $basePath . '/dev_config/rdr_key.json';
+            if (!file_exists($keyFile)) {
+                throw new \Exception("Couldn't find $keyFile");
+            }
+            putenv('GOOGLE_APPLICATION_CREDENTIALS=' . $keyFile);
+
+            $googleClient = new GoogleClient();
+            $googleClient->useApplicationDefaultCredentials();
```
jt2k, line 28: I don't think you need this. The secret manager client is part of the Google Cloud API library and not related to this class.

Also, what do you think about using the same approach as the `GcsBucketService` class where it looks at the environment to make sure it's local before checking for file in dev_config? And you can specify the key path using the `keyFilePath` config option.

jt2k (approved):

PR #1076, opened 2022-03-24
[HPRO-1060] Incentive tracking

shyamnaru:
| Q                   | A
| Bug fix?            | ❌ 
| New feature?        | ✅ 
| Database migration? | ❌ 
| New configuration?  | ❌ 
| Composer updates?   |❌ 
| NPM updates?        | ❌ 
| Jira Ticket(s)      | HPRO-1060


src/Controller/ParticipantDetailsController.php
```diff
@@ -231,4 +262,56 @@ public function participantConsent(
...
+        $id,
+        $incentiveId,
+        Request $request,
+        LoggerService $loggerService,
+        EntityManagerInterface $em,
+        ParticipantSummaryService $participantSummaryService,
+        SiteService $siteService
+    ): Response {
+        $participant = $participantSummaryService->getParticipantById($id);
+        $incentive = $incentiveId ? $em->getRepository(Incentive::class)->find($incentiveId) : null;
```
jt2k, line 279: Shouldn't this also check that the existing incentive is associated with the participant?

src/Controller/ParticipantDetailsController.php
```diff
@@ -231,4 +262,56 @@ public function participantConsent(
...
+        $incentiveForm = $this->createForm(IncentiveType::class, $incentive, ['action' => $request->getRequestUri(), 'require_notes' => true]);
+        $incentiveForm->handleRequest($request);
+        if ($incentiveForm->isSubmitted()) {
+            if ($incentiveForm->isValid()) {
+                $userRepository = $em->getRepository(User::class);
+                $now = new \DateTime();
+                $incentive = $incentiveForm->getData();
+                $incentive->setParticipantId($id);
+                $user = $userRepository->find($this->getUser()->getId());
```
jt2k, line 288: Do we not already have a method to get the current user's user entity? I remember that this isn't as straightforward as it should be because the token user isn't the user entity, but we should probably centralize this somewhere.

src/Entity/Incentive.php
```diff
@@ -0,0 +1,365 @@
...
+
+    /**
+     * @ORM\Column(type="string", length=50)
+     */
+    private $incentiveAmount;
+
+    /**
+     * @ORM\Column(type="integer", nullable=true)
+     */
+    private $otherIncentiveAmount;
```
jt2k, line 88: I understand that this is following the other "other" patterns, but I'm wondering if this should be different. And maybe we talked about this a while ago but I didn't fully understand the situation here. It seems like the incentive amount should always be a number, and not only when "other" is selected. We can discuss more on Monday.

src/Repository/IncentiveRepository.php
```diff
@@ -0,0 +1,21 @@
...
+use Doctrine\Bundle\DoctrineBundle\Repository\ServiceEntityRepository;
+use Doctrine\Persistence\ManagerRegistry;
+
+/**
+ * @method Incentive|null find($id, $lockMode = null, $lockVersion = null)
+ * @method Incentive|null findOneBy(array $criteria, array $orderBy = null)
+ * @method Incentive[]    findAll()
+ * @method Incentive[]    findBy(array $criteria, array $orderBy = null, $limit = null, $offset = null)
+ */
+class IncentiveRepository extends ServiceEntityRepository
```
jt2k, line 15: If there aren't any repository methods, can we remove this file (and the reference to it in the entity)?

templates/participant/details.html.twig
```diff
@@ -85,15 +88,129 @@
...
-                {% endif %}
             </ul>
             <br />
             <div class="tab-content">
-                <div role="tabpanel" class="tab-pane active" id="inperson">
+                <div role="tabpanel" class="tab-pane active" id="on_site_details">
+                    <div class="row">
+                        {% if not isDVType %}
+                            <div class="col-md-6">
+                                {% if canViewPatientStatus %}
```
jt2k, line 98: I like how you put the incentive panel in a separate view partial. Should we do the same for patient status?

web/assets/css/app.css
```diff
@@ -327,6 +327,32 @@ label.label-normal {
     color: #337ab7;
 }
 
+#incentive .panel-heading  a:before {
```
jt2k, line 330: What icon is this? Shouldn't we be using font awesome instead?

jt2k (changes requested):
Looking good overall. A few suggestions and questions.

jt2k (approved):

PR #1079, opened 2022-03-31
[HPRO-1066] Send incentives to RDR

shyamnaru:
| Q                   | A
| Bug fix?            | ❌ 
| New feature?        | ✅
| Database migration? | ❌ 
| New configuration?  | ❌ 
| Composer updates?   | ❌ 
| NPM updates?        | ❌ 
| Jira Ticket(s)      | HPRO-1066


src/Service/IncentiveService.php
```diff
@@ -0,0 +1,133 @@
...
+        LoggerService $loggerService
+    ) {
+        $this->rdrApiService = $rdrApiService;
+        $this->siteService = $siteService;
+        $this->userService = $userService;
+        $this->em = $em;
+        $this->loggerService = $loggerService;
+    }
+
+    public function getRdrObject($incentive, $type = 'create')
```
jt2k, line 30: Instead of passing around the strings 'create', 'amend', and 'cancel', can you use class constants?

jt2k: Can you add tests for this method?

src/Service/IncentiveService.php
```diff
@@ -0,0 +1,133 @@
...
+        }
+        return $obj;
+    }
+
+    public function createIncentive($participantId, $incentiveForm)
+    {
+        $incentive = $incentiveForm->getData();
+        if ($incentive->getIncentiveAmount() === 'other') {
+            $incentive->setIncentiveAmount($incentiveForm['other_incentive_amount']->getData());
+        }
```
jt2k, line 62: Can this logic be implemented as a data mapper in the form?

src/Service/IncentiveService.php
```diff
@@ -0,0 +1,133 @@
...
+            if ($incentive->getGiftCardType()) {
+                $obj->giftcardType = $incentive->getGiftCardType();
+            }
+            $obj->amount = $incentive->getIncentiveAmount();
+            $obj->notes = $incentive->getNotes();
+        }
+        return $obj;
+    }
+
+    public function createIncentive($participantId, $incentiveForm)
```
jt2k, line 57: It would be good to have tests for these other methods too. It will be a bit more complicated than the getRdrObject method, but I think it can be done by mocking the RDR, user, and site services.

jt2k (changes requested):

src/Service/IncentiveService.php
shyamnaru: Do you have an example of how to implement that using data mapper?

jt2k: I might have gotten data mappers and data transformers confused.

Reading through the documentation, I _think_ a data transformer makes sense here: https://symfony.com/doc/5.4/form/data_transformers.html

And here's the section in the data mapper documentation about the difference between mappers and transformers: https://symfony.com/doc/5.4/form/data_mappers.html#the-difference-between-data-transformers-and-mappers

I don't have a specific example, but the documentation provides a few. If this doesn't make sense in this case, that's ok. I was just looking for a way to not have to repeat this logic in both the create and amend methods.

tests/Service/IncentiveServiceTest.php
```diff
@@ -0,0 +1,66 @@
...
+        self::assertEquals(1, $rdrObject->incentiveId);
+
+        $incentive = $this->createIncentive(Incentive::CANCEL);
+        $rdrObject = $this->service->getRdrObject($incentive, Incentive::CANCEL);
+        self::assertEquals('test@example.com', $rdrObject->cancelledBy);
+        self::assertEquals(true, $rdrObject->cancel);
+        self::assertEquals(1, $rdrObject->incentiveId);
+
+
+
```
jt2k, line 48: Can you remove these extra line breaks?

tests/Service/IncentiveServiceTest.php
```diff
@@ -0,0 +1,66 @@
...
+    protected $id;
+
+    public function setUp(): void
+    {
+        parent::setUp();
+        $this->login('test@example.com', ['hpo-site-test'], 'America/Chicago');
+        $siteService = static::$container->get(SiteService::class);
+        $siteService->switchSite('hpo-site-test' . '@' . self::GROUP_DOMAIN);
+        $this->service = static::$container->get(IncentiveService::class);
+
```
jt2k, line 21: Can you remove this extra line break?

jt2k (approved):
Looks good, other than some extra line breaks in the test class.

PR #1093, opened 2022-04-15
Be/phpstan 2 secureuser

thornview:
| Q                   | A
| Bug fix?            | ✅ 
| New feature?        | ❌ 
| Database migration? |❌ 
| New configuration?  | ❌ 
| Composer updates?   | ❌ 
| NPM updates?        | ❌ 
| Jira Ticket(s)      | HPRO-... 

### Summary
I started addressing PHPStan Level 2 issues and am beginning with this PR.  

A number of controller methods were calling `$this->getUser()` and receiving a `UserInterface` object in response.  In most cases, the code assumes that the `UserInterface` object is actually an instance of `App\Security\User` and proceeds to call methods specific to that class (like `getTimeZone`).  

PHPStan was flagging these instances as errors because the `UserInterface` being promised does not include the `App\Security\User` methods.  

I've addressed this issue by creating a `getSecureUser()` method in the `BaseController` which promises to return an instance of `App\Security\User`.  I then updated all of the controllers to extend the `BaseController`.  

I'm not sure that's the cleanest way to handle the issue, but this at least centralizes the calls so that we can modify the code in single place if we want to take a different approach.  

### Instructions for testing  
This update touches a lot of routes.  It would be a good idea to drive through the application and make sure things still work.


thornview:
@jt2k Refactoring `loadUserByUsername` to return `App\Security\User` did not resolve the PHPStan issues in the controllers, although it may be useful elsewhere.  In the controllers the method `this->getUser()` maps to the `AbstractController`, which gets the `User` from `TokenStorage` and promises to return a `UserInterface`.

thornview:
@jt2k Refactoring `loadUserByUsername` to return `App\Security\User` did not resolve the PHPStan issues in the controllers, although it may be useful elsewhere.  In the controllers the method `this->getUser()` maps to the `AbstractController`, which gets the `User` from `TokenStorage` and promises to return a `UserInterface`.

shyamnaru (approved):

PR #1096, opened 2022-04-20
[HPRO-1017] Implement id verification

shyamnaru:
| Q                   | A
| Target Release?     | 2.8.1 
| Bug fix?            | ❌ 
| New feature?        | ✅
| Database migration? | ❌ 
| New configuration?  | ❌ 
| Composer updates?   | ❌ 
| NPM updates?        | ❌ 
| Jira Ticket(s)      | HPRO-1017

### DIFF
--- a/src/Controller/ParticipantDetailsController.php
+++ b/src/Controller/ParticipantDetailsController.php
@@ -182,13 +182,12 @@ class ParticipantDetailsController extends BaseController
             if ($idVerificationForm->isValid()) {
                 if ($idVerificationService->createIdVerification($id, $idVerificationForm->getData())) {
                     $this->addFlash('success', 'ID Verification Saved');
-                } else {
-                    $this->addFlash('error', 'Error saving id verification . Please try again');
+                    return $this->redirectToRoute("participant", ['id' => $id]);
                 }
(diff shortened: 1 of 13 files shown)


src/Controller/ParticipantDetailsController.php
```diff
@@ -172,6 +175,23 @@ public function participantDetailsAction(
...
+            if ($idVerificationForm->isValid()) {
+                if ($idVerificationService->createIdVerification($id, $idVerificationForm->getData())) {
+                    $this->addFlash('success', 'ID Verification Saved');
+                } else {
+                    $this->addFlash('error', 'Error saving id verification . Please try again');
+                }
+            } else {
+                $this->addFlash('error', 'Invalid form');
+            }
+            return $this->redirectToRoute("participant", ['id' => $id]);
```
jt2k, line 191: Not sure if there's something different with the multiple forms on this page, but wouldn't we only want to redirect if the form submission was successful? Right now, it would just display a generic "invalid form" and the form would be cleared, instead of displaying the form with their previous submission and inline errors.

jt2k (approved):

PR #1114, opened 2022-05-12
[HPRO-1099] Implement incentive import

shyamnaru:
| Q                   | A
| Target Release?     | /Next available 
| Bug fix?            | ❌ 
| New feature?        | ✅ 
| Database migration? | ❌ 
| New configuration?  | ❌ 
| Composer updates?   | ❌ 
| NPM updates?        | ❌ 
| Jira Ticket(s)      | HPRO-1099


src/Controller/IncentiveImportController.php
```diff
@@ -0,0 +1,136 @@
...
+        LoggerService $loggerService,
+        IncentiveImportService $incentiveImportService
+    ) {
+        $form = $this->createForm(IncentiveImportFormType::class);
+        $form->handleRequest($request);
+        if ($form->isSubmitted()) {
+            $file = $form['incentive_csv']->getData();
+            $fileName = $file->getClientOriginalName();
+            $incentives = [];
+            $incentiveImportService->extractCsvFileData($file, $form, $incentives);
```
jt2k, line 41: Instead of passing in an empty array by reference that is modified inside the function, can we just have this method return the resulting array?

src/Controller/IncentiveImportController.php
```diff
@@ -0,0 +1,136 @@
...
+                        $this->em->persist($incentiveImportRow);
+                        if (($key % $batchSize) === 0) {
+                            $this->em->flush();
+                            $this->em->clear(IncentiveImportRow::class);
+                        }
+                    }
+                    $this->em->flush();
+                    $id = $incentiveImport->getId();
+                    $loggerService->log(Log::INCENTIVE_IMPORT_ADD, $id);
+                    $this->em->clear();
```
jt2k, line 76: Can this be moved to the import service?

src/Controller/IncentiveImportController.php
```diff
@@ -0,0 +1,136 @@
...
+        $form->handleRequest($request);
+        if ($form->isSubmitted() && $form->isValid()) {
+            /** @var SubmitButton $confirmButton */
+            $confirmButton = $form->get('Confirm');
+            if ($confirmButton->isClicked()) {
+                // Update confirm status
+                $incentiveImport->setConfirm(1);
+                $this->em->flush();
+                $loggerService->log(Log::INCENTIVE_IMPORT_EDIT, $incentiveImport->getId());
+                $this->em->clear();
```
jt2k, line 112: Why does the entity manager need to be cleared here?

src/Entity/IncentiveImport.php
```diff
@@ -0,0 +1,169 @@
...
+
+    public function removeIncentiveImportRow(IncentiveImportRow $IncentiveImportRow): self
+    {
+        if ($this->IncentiveImportRows->contains($IncentiveImportRow)) {
+            $this->IncentiveImportRows->removeElement($IncentiveImportRow);
+            // set the owning side to null (unless already changed)
+            if ($IncentiveImportRow->getImport()->getId() === $this->getId()) {
+                $IncentiveImportRow->setImport(null);
+            }
+        }
```
jt2k, line 165: Why is the `$IncentiveImportRow` variable capitalized?

src/Repository/IncentiveImportRepository.php
```diff
@@ -0,0 +1,50 @@
+<?php
```
jt2k, line 1: As [discussed previously](https://github.com/all-of-us/healthpro/pull/1111/files#r869516860), if we don't have any custom repository methods, let's remove the empty repository classes (and update the entity definition to remove the `repositoryClass` annotation attribute)

src/Service/IncentiveImportService.php
```diff
@@ -0,0 +1,116 @@
+<?php
+
+namespace App\Service;
+
+use App\Entity\Incentive;
+use Doctrine\ORM\EntityManagerInterface;
+use Symfony\Component\DependencyInjection\ParameterBag\ParameterBagInterface;
+use Symfony\Component\Form\FormError;
+
+class IncentiveImportService
```
jt2k, line 10: Tests for this class?

jt2k (changes requested):

src/Controller/IncentiveImportController.php
shyamnaru: We are passing both $form & $incentives as reference. $form is passed to get the error messages. That's why I passed both as reference.

src/Entity/IncentiveImport.php
shyamnaru: I think this is because it's mapped to IncentiveImportRow entity.

src/Controller/IncentiveImportController.php
jt2k: What I didn't understand is why the need to create the $incentives array outside the method at all, instead of just returning and assigning it?

```php
$incentives = $incentiveImportService->extractCsvFileData($file, $form);
```

and then for the `extractCsvFileData` method:

```php
public function extractCsvFileData($file, $form): void
{
    $incentives = [];
    // ...
    return $incentives;
}
```

Also, unless I'm misunderstanding something, since the form variable is an object, isn't that passed by reference anyways? I don't think you need the `&` in the method signature.

src/Entity/IncentiveImport.php
jt2k: Aren't all of our variable names always lowercase? I don't think we ever treat instances of entity objects any differently.

shyamnaru: This code is generated by maker bundle, I see that in other places too https://github.com/all-of-us/healthpro/blob/develop/src/Entity/PatientStatusImport.php#L163. Maybe we should change to lowercase?

thornview: I thought the PSR standards addressed the question of variable capitalization but it turns out they purposely avoid it.  Nevertheless, let's make all method and variable names start with a lowercase letter, while class names will start with a capital letter.

thornview: Turns out this capitalization rule is in the [Symfony Coding Standards](https://symfony.com/doc/current/contributing/code/standards.html#naming-conventions).  That explains why I couldn't find them in the PSR.

src/Controller/IncentiveImportController.php
```diff
@@ -16,9 +17,12 @@
 
 class IncentiveImportController extends BaseController
 {
-    public function __construct(EntityManagerInterface $em)
+    public function __construct(EntityManagerInterface $em, EnvironmentService $env)
     {
         parent::__construct($em);
+        if (!$env->isLocal()) {
```
jt2k, line 23: Should we use a feature flag configuration instead?

tests/Service/IncentiveImportServiceTest.php
```diff
@@ -39,4 +40,32 @@ public function emailDataProvider()
...
+     */
+    public function testExtractCsvFileData($fileName, $isValid)
+    {
+        $form = static::$container->get('form.factory')->create(IncentiveImportFormType::class, null, ['csrf_protection' => false]);
+        $form->submit([
+            'incentive_csv' => $this->createUploadedFile($fileName)
+        ]);
+        $file = $form['incentive_csv']->getData();
+        $this->service->extractCsvFileData($file, $form);
+        $this->assertEquals($form->isValid(), $isValid);
```
jt2k, line 55: Could you also check the results of the `extractCsvFileData` method when it is valid? At least a check for the expected number of rows?

src/Controller/IncentiveImportController.php
shyamnaru: I am not sure if we want to use a feature flag since this is not completely done?

jt2k (approved):

PR #1116, opened 2022-05-23
[HPRO-1113] Add Deceased Reporting to Support Center View

shyamzcode:
| Q                   | A
| Bug fix?            | ❌ 
| New feature?        | ✅
| Database migration? | ❌ 
| New configuration?  | ❌ 
| Composer updates?   | ❌ 
| NPM updates?        | ❌ 
| Jira Ticket(s)      | HPRO-1113


web/assets/js/views/DeceasedReport.js
```diff
@@ -0,0 +1,13 @@
+$(document).ready(function () {
+    $('input:radio').on('change', function () {
```
jt2k, line 2: Instead of listening for any radio button change, can we add something to limit the scope here? Maybe add an id to row div and target input:radio inside that div?

templates/deceasedreports/check.html.twig
```diff
@@ -0,0 +1,32 @@
...
+            <div style="display: none;" class="alert alert-danger" id="deceased_check_warning">
+                Please follow the withdrawal workflow for this participant.
+            </div>
+        </div>
+    </div>
+{% endblock %}
+
+{% block pagejs %}
+    {{ encore_entry_script_tags('deceased-report') }}
+{% endblock %}
```
jt2k, line 32: I think we want our twig files to end with a new line like our PHP files for consistency.

jt2k (changes requested):
Looks good. Just two minor things.

jt2k (approved):

PR #1119, opened 2022-05-25
[HPRO-1100] Incentive tracking view import details

shyamnaru:
| Q                   | A
| Bug fix?            | ❌ 
| New feature?        | ✅ 
| Database migration? | ❌ 
| New configuration?  | ❌ 
| Composer updates?   | ❌ 
| NPM updates?        | ❌ 
| Jira Ticket(s)      | HPRO-1100

### DIFF
--- a/src/Controller/IncentiveImportController.php
+++ b/src/Controller/IncentiveImportController.php
@@ -11,6 +11,7 @@ use App\Service\LoggerService;
 use Doctrine\ORM\EntityManagerInterface;
 use Symfony\Component\Form\FormError;
 use Symfony\Component\Form\SubmitButton;
+use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\Routing\Annotation\Route;
 use App\Audit\Log;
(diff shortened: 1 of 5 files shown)


src/Controller/IncentiveImportController.php
```diff
@@ -95,8 +96,25 @@ public function incentiveImportConfirmation(int $id, Request $request, LoggerSer
...
+            'user' => $this->getUserEntity(),
+            'confirm' => 1
+        ]);
+        if (empty($incentiveImport)) {
+            throw $this->createNotFoundException('Page Not Found!');
+        }
+        //For ajax requests
+        if ($request->isXmlHttpRequest()) {
+            $params = $request->request->all();
+            $incentiveImportRows = $incentiveImport->getIncentiveImportRows()->slice($params['start'], $params['length']);
```
jt2k, line 112: Is there a limit on the number of rows that can be imported at once? If not, we may want to consider paginating on the database side, instead of hydrating the entire collection on every request.

jt2k (approved):
Looks good. I have a question about how well this scales, but maybe this does not need to be optimized.

src/Controller/IncentiveImportController.php
shyamnaru: Yes, there is a 5000 row limit for the CSV import.

PR #1124, opened 2022-06-03
[HPRO-1102] Incentive import cron job

shyamnaru:
| Q                   | A
| Target Release?     | 2.8.4 
| Bug fix?            | ❌ 
| New feature?        | ✅/
| Database migration? | ❌ 
| New configuration?  | ❌ 
| Composer updates?   | ❌ 
| NPM updates?        | ❌ 
| Jira Ticket(s)      | HPRO-1102

### DIFF
--- a/src/Repository/IncentiveImportRowRepository.php
+++ b/src/Repository/IncentiveImportRowRepository.php
@@ -1,0 +1,41 @@
+<?php
+
+namespace App\Repository;
+
+use App\Entity\IncentiveImportRow;
+use Doctrine\Bundle\DoctrineBundle\Repository\ServiceEntityRepository;
+use Doctrine\Persistence\ManagerRegistry;
+
+/**
+ * @method IncentiveImportRow|null find($id, $lockMode = null, $lockVersion = null)
+ * @method IncentiveImportRow|null findOneBy(array $criteria, array $orderBy = null)
+ * @method IncentiveImportRow[]    findAll()
+ * @method IncentiveImportRow[]    findBy(array $criteria, array $orderBy = null, $limit = null, $offset = null)
+ */
+class IncentiveImportRowRepository extends ServiceEntityRepository
+{
+    public function __construct(ManagerRegistry $registry)
+    {
+        parent::__construct($registry, IncentiveImportRow::class);
+    }
+
+    public function getIncentiveImportRows($limit): array
+    {
+        $query = "
+            SELECT iir.*,
+                   ii.site,
+                   ii.created_ts
+            FROM incentive_import_row iir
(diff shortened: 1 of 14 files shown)


src/Repository/IncentiveImportRowRepository.php
```diff
@@ -0,0 +1,49 @@
...
+class IncentiveImportRowRepository extends ServiceEntityRepository
+{
+    public function __construct(ManagerRegistry $registry)
+    {
+        parent::__construct($registry, IncentiveImportRow::class);
+    }
+
+    public function getIncentiveImportRows($limit): array
+    {
+        $query = "
```
jt2k, line 24: Can we use the query builder here instead of raw SQL?

src/Repository/IncentiveImportRowRepository.php
```diff
@@ -0,0 +1,49 @@
...
+ * @method IncentiveImportRow[]    findBy(array $criteria, array $orderBy = null, $limit = null, $offset = null)
+ */
+class IncentiveImportRowRepository extends ServiceEntityRepository
+{
+    public function __construct(ManagerRegistry $registry)
+    {
+        parent::__construct($registry, IncentiveImportRow::class);
+    }
+
+    public function getIncentiveImportRows($limit): array
```
jt2k, line 22:
```suggestion
    public function getIncentiveImportRows(int $limit): array
```

src/Repository/IncentiveImportRowRepository.php
```diff
@@ -0,0 +1,49 @@
...
+        }
+        return $this->getEntityManager()->getConnection()->fetchAll($query, [
+            'confirm' => 1,
+            'rdrStatus' => 0
+        ]);
+    }
+
+    public function deleteUnconfirmedImportData($date): void
+    {
+        $query = "DELETE iir FROM incentive_import_row iir inner join incentive_import ii on iir.import_id = ii.id where ii.created_ts < :date and ii.confirm = :confirm";
```
jt2k, line 44: DQL or query builder here, too?

src/Service/IncentiveImportService.php
```diff
@@ -190,4 +209,157 @@ public function getAjaxData($incentiveImport, $incentiveImportRows)
         }
         return $rows;
     }
+
+    public function sendIncentive($participantId, $incentive, $user): bool
```
jt2k, line 213: Unless I missed something, it looks like this is only used from inside the `sendIncentivesToRdr` method below. Can this be changed to a private method?

Or, could the IncentiveService be used here? It looks like the method there is coupled to the form, but it might be worth having just one place in our code where we send an incentive to the RDR.

jt2k (changes requested):

jt2k (approved):

PR #1148, opened 2022-07-14
[HPRO-1101] Add mfa check while adding users to a group

shyamzcode:
| Q                   | A
| Bug fix?            | ❌ 
| New feature?        | ✅ 
| Database migration? | ❌ 
| New configuration?  | ❌ 
| Composer updates?   | ❌ 
| NPM updates?        | ❌ 
| Jira Ticket(s)      | HPRO-1101

### DIFF
--- a/src/Controller/AccessManagementController.php
+++ b/src/Controller/AccessManagementController.php
@@ -21,7 +21,7 @@ use Symfony\Component\Routing\Annotation\Route;
 class AccessManagementController extends BaseController
 {
     public const MEMBER_DOMAIN = '@pmi-ops.org';
-    public const RESET_PASSWORD_URL = 'https://admin.google.com';
+    public const GOOGLE_ADMIN_URL = 'https://admin.google.com';
 
     private $googleGroupsService;
     private $loggerService;
@@ -70,7 +70,7 @@ class AccessManagementController extends BaseController
         return $this->render('accessmanagement/group-members.html.twig', [
             'group' => $group,
             'members' => $members,
-            'resetPasswordUrl' => self::RESET_PASSWORD_URL
+            'resetPasswordUrl' => self::GOOGLE_ADMIN_URL
         ]);
     }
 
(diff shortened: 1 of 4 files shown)


src/Service/GoogleGroupsService.php
```diff
@@ -224,4 +225,19 @@ public function removeMember(string $groupEmail, string $email)
...
+        $emails = [];
+        foreach ($users as $user) {
+            $emails[] = $user->email;
+        }
+        return $emails;
+    }
+
+    public function isMfaGroupUser($email): bool
+    {
+        return in_array($email, $this->getMfaGroupUserEmails());
```
jt2k, line 241: Instead of retrieving all members in the group and looking for the one user, I think it would be better to look at all of the user's groups to see if they are a member of the exception group.

I think the number of new users being onboarded and belonging to the 2fa exception group could grow as the AoU program grows, while the maximum number of sites that a single user belongs to would be fairly static.

src/Controller/AccessManagementController.php
```diff
@@ -89,16 +89,20 @@ public function member($groupId, Request $request)
...
+                            $this->addFlash('success', $result['message']);
+                            $this->loggerService->log(Log::GROUP_MEMBER_ADD, [
+                                'member' => $email,
+                                'result' => 'success'
+                            ]);
+                            return $this->redirectToRoute('access_manage_user_group', ['groupId' => $groupId]);
+                        }
+                        $errorMessage = isset($result['code']) && $result['code'] === 409 ? 'Member already exists.' : 'Error occurred. Please try again.';
+                    } else {
+                        $errorMessage = ' User cannot be added until 2FA is enabled. Please check the user’s 2FA status in the Admin Console before contacting drcsupport@vumc.org for assistance.';
```
jt2k, line 104:
```suggestion
                        $errorMessage = 'User cannot be added until 2FA is enabled. Please check the user’s 2FA status in the Admin Console before contacting drcsupport@vumc.org for assistance.';
```

jt2k (changes requested):

jt2k (approved):

PR #1153, opened 2022-07-20
[HPRO-1094] Create admin panel feature notifications

shyamzcode:
| Q                   | A
| Bug fix?            | ❌ 
| New feature?        | ✅ 
| Database migration? |❌ 
| New configuration?  | ❌ 
| Composer updates?   | ❌ 
| NPM updates?        | ❌ 
| Jira Ticket(s)      | HPRO-1094


src/Form/FeatureNotificationType.php
```diff
@@ -0,0 +1,116 @@
...
+                ]
+            ])
+            ->add('url', Type\TextType::class, [
+                'label' => 'Direct User To (URL Pattern)',
+                'required' => false,
+                'constraints' => [
+                    new Constraints\Type('string'),
+                    new Constraints\Regex('/^[a-zA-Z0-9_\-\/\*]+$/') // valid URL, with asterisks
+                ]
+            ])
```
jt2k, line 61: Shouldn't this just be a URL, and not a URL pattern? I would suggest updating the label and removing the asterisk from the regex.

src/Form/FeatureNotificationType.php
```diff
@@ -0,0 +1,116 @@
...
+            ->add('message', Type\TextareaType::class, [
+                'required' => true,
+                'attr' => ['rows' => 4],
+                'constraints' => [
+                    new Constraints\NotBlank(),
+                    new Constraints\Type('string')
+                ]
+            ])
+            ->add('new_feature_status', Type\CheckboxType::class, [
```
jt2k, line 44: Do we need a separate field for this? Could it just be derived from whether or not there is a URL?

shyamzcode: I had the same doubt when working on this but I just implemented as it's mentioned in the mockup. I'll check with Gwen and see what she says.

src/Form/FeatureNotificationType.php
shyamzcode: Will check with Gwen as see what she says.

web/assets/js/views/FeatureNotification.js
```diff
@@ -0,0 +1,19 @@
+$(document).ready(function () {
+    $('#feature_notification_start_ts, #feature_notification_end_ts').pmiDateTimePicker();
+    $('#feature_notification_url').dropdownOther({
+        'All': '/*',
+        'Home Page': '/',
+        'Work Queue': '/workqueue/',
+        'Biobank Order Pages': '/participant/*/order/*',
+        'Physical Measurements': '/participant/*/measurements*',
+        'Management Tools': '/access/manage/dashboard',
+        'User Management': '/access/manage/user/groups'
```
jt2k, line 10: These need to be updated. Does Gwen have a list of pages she wants as dropdown options?

jt2k (changes requested):
The new changes look good. The only thing left is the URL dropdown options. The current ones are still the patterns from the page notices.

web/assets/js/views/FeatureNotification.js
shyamzcode: Definitely, we need to remove the ones which has asterisk. Also, commented in the ticket asking Gwen to get the list.

PR #1166, opened 2022-08-09
[HPRO-1115] New Feature Notifications - Customize View Per User

shyamzcode:
| Q                   | A
| Bug fix?            | ❌ 
| New feature?        | ✅ 
| Database migration? | ❌ 
| New configuration?  | ❌ 
| Composer updates?   | ❌ 
| NPM updates?        | ❌ 
| Jira Ticket(s)      | HPRO-1115

### DIFF
--- a/templates/partials/notifications-dropdown.html.twig
+++ b/templates/partials/notifications-dropdown.html.twig
@@ -17,7 +17,9 @@
     {% for notification in global_notifications %}
         <div class="row">
             <div class="col-sm-1">
-                <i class='fa fa-circle text-danger'></i>
+                {% if notification.id not in user_notification_ids %}
+                    <i class='fa fa-circle text-danger'></i>
+                {% endif %}
             </div>
             <div class="col-sm-9">
                 <p><a data-href="{{ path('notification_details', {id: notification.id}) }}" class="notification-title">{{ notification.title }}</a></p>
--- a/templates/partials/notifications-dropdown.html.twig
+++ b/templates/partials/notifications-dropdown.html.twig
@@ -9,14 +9,15 @@
             </strong>
         </div>
         <div class="col-sm-3 col-sm-offset-5">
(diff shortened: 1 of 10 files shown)


templates/partials/notifications-dropdown.html.twig
```diff
@@ -9,15 +9,17 @@
             </strong>
         </div>
         <div class="col-sm-3 col-sm-offset-5">
-            <a>Mark all as read</a>
+            <a href="{{ path('notifications_mark_read', { return: app.request.pathinfo }) }}">Mark all as read</a>
```
jt2k, line 12: Can you implement this as an AJAX request (and then clear the navbar badge when it completes)?

I know that would add a bit more front-end complexity, but it would prevent the page from having to be refreshed. The current behavior might be unexpected for the user. 

If you do end up sticking with this request and return redirect, you'll need to check the return path to make sure it is a relative URL. On the settings page, we added that check, to prevent the chance of a user getting tricked into clicking a link with a malicious return URL.

jt2k (changes requested):
This looks good! I do have a concern about the mark all as read functionality, though. See inline comment.

templates/partials/notifications-dropdown.html.twig
```diff
@@ -9,14 +9,15 @@
             </strong>
         </div>
         <div class="col-sm-3 col-sm-offset-5">
-            <a href="{{ path('notifications_mark_read', { return: app.request.pathinfo }) }}">Mark all as read</a>
+            <a data-href="{{ path('notifications_mark_read') }}" id="notifications_mark_all_read"
```
jt2k, line 12: Since this isn't a link anymore, can you use another element instead of `<a>`?

jt2k (approved):
Thanks for the changes! One request to change the element tag, but otherwise, approved.

jt2k (approved):

PR #1205, opened 2022-10-05
[HPRO-1046] Edited templates and added method for sop titles

Ehaic:
| Q                   | A
| Bug fix?            | ❌ 
| New feature?        | ✅ 
| Database migration? | ❌ 
| New configuration?  | ❌ 
| Composer updates?   | ❌ 
| NPM updates?        | ❌ 
| Jira Ticket(s)      | HPRO-1046 

### Summary

Adds links to SOP's for measurement processes. Clicking links open SOP's in new tab.

### Screenshots 

image

image

### DIFF
--- a/tests/Service/HelpServiceTest.php
+++ b/tests/Service/HelpServiceTest.php
@@ -1,0 +1,18 @@
+<?php
+namespace App\Tests\Service;
+
+use App\Service\HelpService;
+
+class HelpServiceTest extends ServiceTestCase {
+    protected $helpService;
+
+    public function testGetDocumentTitlesList(): void
+    {
+        $this->helpService = static::getContainer()->get(HelpService::class);
+        $documentList = $this->helpService->getDocumentTitlesList();
+        self::assertIsArray($documentList);
+        self::assertGreaterThan(0, count($documentList));
+        self::assertArrayHasKey('SOP-014', $documentList);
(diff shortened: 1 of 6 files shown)


tests/Service/HelpServiceTest.php
```diff
@@ -0,0 +1,18 @@
...
+namespace App\Tests\Service;
+
+use App\Service\HelpService;
+
+class HelpServiceTest extends ServiceTestCase {
+    protected $helpService;
+
+    public function testGetDocumentTitlesList(): void
+    {
+        $this->helpService = static::getContainer()->get(HelpService::class);
```
jt2k, line 11: A couple minor things here:

1. We follow PSR-12 for code style, so the opening brace for a class should be on a new line. We have an automated check for this using PHP CS Fixer, but it looks like our configuration is only looking at the `src` directory, so our files currently aren't being checked. We should look into updating that sometime.
2. Since you're getting the HelpService from inside this test method, there's no reason to assign it to a class variable. I would suggest just setting it to a local variable to use in the test, or, if you prefer having it available in the class for other future tests, then you could add a `setUp` method and set it there, which will run before every test.

jt2k (approved):
Looks good! A couple minor suggestions.

jt2k (approved):

PR #1248, opened 2022-12-07
[HPRO-1184] User Management tools.

Ehaic:
| Q                   | A
| Target Release?     | 3.0.0
| Bug fix?            | ❌ 
| New feature?        | ✅ 
| Database migration? | ❌ 
| New configuration?  | ❌ 
| Composer updates?   | ❌ 
| NPM updates?        | ❌ 
| Jira Ticket(s)      | HPRO-1184 

### Screenshots 

image
image


src/Controller/AccessManagementController.php
```diff
@@ -49,15 +58,19 @@ public function index()
...
+        return $this->render($this->contextTemplate->GetProgramTemplate('accessmanagement/groups.html.twig'));
     }
 
     /**
      * @Route("/user/group/{groupId}", name="access_manage_user_group")
      */
     public function userGroup($groupId): Response
     {
-        $group = $this->getSecurityUser()->getGroupFromId($groupId);
+        if ($this->requestStack->getSession()->get('program') === 'hpo') {
```
thornview, line 69: Instead of using magic strings "hpo" and "nph", let's reference a constant.  Since we're calling this same logic frequently, it would also be a good idea to create `isHealthPro` and `isNph` tests in the `BaseController` or as static methods in some other service.

thornview (changes requested):

shyamzcode:
If a user belongs to both HealthPro and NPH sites I think we shouldn't display the Management Tool in the menu until the user select a program. Currently, when a user clicks on the Management Tool with out selecting the program it takes to the HealthPro user management page. Rest looks good!

Screen Shot 2022-12-09 at 3 29 27 PM

shyamzcode (approved):

src/Controller/AccessManagementController.php
Ehaic: Will make a seperate issue to implement this.

thornview (approved):

PR #1249, opened 2022-12-07
[HPRO-1286] Sample finalize implement validations

shyamzcode:
| Q                   | A
| Target Release?     | 3.0.0 
| Bug fix?            | ❌ 
| New feature?        | ✅ 
| Database migration? | ❌ 
| New configuration?  | ❌ 
| Composer updates?   | ❌ 
| NPM updates?        | ❌ 
| Jira Ticket(s)      | HPRO-1286 

### Screenshots 
image

image

### DIFF
--- a/src/Service/Nph/NphOrderService.php
+++ b/src/Service/Nph/NphOrderService.php
@@ -403,9 +403,11 @@ class NphOrderService
     {
         $aliquots = $this->getAliquots($sampleCode);
         foreach ($aliquots as $aliquotCode => $aliquot) {
-            foreach ($formData[$aliquotCode] as $aliquotId) {
-                if ($aliquotId) {
-                    return true;
+            if (isset($formData[$aliquotCode])) {
+                foreach ($formData[$aliquotCode] as $aliquotId) {
+                    if ($aliquotId) {
+                        return true;
+                    }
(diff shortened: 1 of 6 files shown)


src/Service/Nph/NphOrderService.php
```diff
@@ -399,6 +397,21 @@ private function generateOrderSummaryArray(array $nphOrder): array
...
+    {
+        $aliquots = $this->getAliquots($sampleCode);
+        foreach (array_keys($aliquots) as $aliquotCode) {
+            if (isset($formData[$aliquotCode])) {
+                foreach ($formData[$aliquotCode] as $aliquotId) {
+                    if ($aliquotId) {
+                        return true;
+                    }
+                }
+            }
```
thornview, line 410: Could this be simplified to: 
```suggestion
            if (isset($formData[$aliquotCode]) && count($formData[$aliquoteCode]) > 0) {
                return true;
            }
```

shyamzcode: @thornview This is how the collection form data is represented. The count will be always > 0, so we need iterate the loop and determine if the field has null value or not.

image

Ehaic (approved):

PR #1261, opened 2022-12-14
[HPRO-1282] Adding Exploring the Mind Consent to Workqueue and Participant Summary

Ehaic:
| Q                   | A
| Bug fix?            | ❌ 
| New feature?        | ✅ 
| Database migration? | ❌ 
| New configuration?  | ❌ 
| Composer updates?   | ❌ 
| NPM updates?        | ❌ 
| Jira Ticket(s)      | HPRO-1282 

### Screenshots 

image
image
image

### DIFF
--- a/templates/participant/details.html.twig
+++ b/templates/participant/details.html.twig
@@ -179,7 +179,9 @@
                                             {% for type, label in digitalHealthSharingTypes %}
                                                 <dd>{{ label }} {{ macros.displayDigitalSharing(participant.digitalHealthSharingStatus, type) }}</dd>
                                             {% endfor %}
-                                        </dl>
+                                            </dl>
+                                            <dt>Exploring the Mind Consent</dt>
+                                            <dd>{{ macros.displayConsentPanelStatus(participant, 'EtMConsent', '', true) }}</dd>
                                     </div>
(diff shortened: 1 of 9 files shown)


templates/participant/details.html.twig
```diff
@@ -179,7 +179,9 @@
                                             {% for type, label in digitalHealthSharingTypes %}
                                                 <dd>{{ label }} {{ macros.displayDigitalSharing(participant.digitalHealthSharingStatus, type) }}</dd>
                                             {% endfor %}
-                                        </dl>
+                                            </dl>
+                                            <dt>Exploring the Mind Consent</dt>
+                                            <dd>{{ macros.displayConsentPanelStatus(participant, 'EtMConsent', '', true) }}</dd>
```
shyamzcode, line 184: This should be consentForEtM instead of EtMConsent. Also, I confirmed with Gwen we don't need to display red x or green check. Could you pass false instead of true so that we don't display those check marks?

shyamzcode (changes requested):
There where few issues and I left few comments. Please review those but rest looks good, thanks!

shyamzcode (approved):

PR #1312, opened 2023-02-03
[HPRO-1341] Validate generate orders page

shyamzcode:
| Q                   | A
| Target Release?     | 3.0.0 
| Bug fix?            | ❌ 
| New feature?        | ✅ 
| Database migration? | ❌ 
| New configuration?  | ❌ 
| Composer updates?   | ❌ 
| NPM updates?        | ❌ 
| Jira Ticket(s)      | HPRO-1341 

### Screenshots 
image
image
image
image
image

### DIFF
--- a/src/Service/Nph/NphOrderService.php
+++ b/src/Service/Nph/NphOrderService.php
@@ -896,20 +896,31 @@ class NphOrderService
 
     public function validateGenerateOrdersData(array $formData): array
     {
-        // For stool kit samples
+        $formErrors = [];
         if (!empty($formData['stoolKit'])) {
+            $hasStoolTube = false;
             foreach ($this->getSamplesByType('stool') as $stoolSample) {
                 if (!empty($formData[$stoolSample])) {
-                    return [];
+                    $hasStoolTube = true;
                 }
             }
-            return [
-                [
+            if ($hasStoolTube === false) {
+                $formErrors[] = [
                     'field' => $this->getSamplesByType('stool')[0],
-                    'errorMessage' => 'Please enter at least one stool tube id'
-                ]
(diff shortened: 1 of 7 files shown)


src/Service/Nph/NphOrderService.php
```diff
@@ -893,4 +893,58 @@ public function getVisitTypes(): array
...
+        if (empty($formData['stoolKit'])) {
+            if ($hasSample === false) {
+                $formErrors[] = [
+                    'field' => 'checkAll',
+                    'message' => 'Please select or enter at least one sample'
+                ];
+                return $formErrors;
+            }
+        }
+        if (!empty($formData['stoolKit'])) {
```
thornview, line 916: It looks like this could be simplified into an `if ... else` 

```suggestion
        if (empty($formData['stoolKit'])) {
            if ($hasSample === false) {
                $formErrors[] = [
                    'field' => 'checkAll',
                    'message' => 'Please select or enter at least one sample'
                ];
                return $formErrors;
            }
        }
       else {
```

web/assets/js/views/NphOrderCreate.js
```diff
@@ -23,7 +24,7 @@ $(document).ready(function () {
                     .each(function () {
                         if ($(this).prop("checked") === true && $(this).prop("disabled") === false) {
                             let sample = $(this).val();
-                            if (sample === "nail") {
+                            if (sample === "NAIL") {
```
thornview, line 27: It would be good to use constants for these magic strings `NAIL` and `STOOL`

thornview (approved):
A couple recommendations for changes, but looks good overall.

PR #1366, opened 2023-03-14
[HPRO-1371] Added requirement of atleast one 500 microliter aliquot.

Ehaic:
| Q                   | A
| Target Release?     | 3.2.0 
| Bug fix?            | ❌ 
| New feature?        | ❌ 
| Database migration? | ❌ 
| New configuration?  | ❌ 
| Composer updates?   | ❌ 
| NPM updates?        | ❌ 
| Jira Ticket(s)      | HPRO-1371 

### Screenshots

### DIFF
--- a/src/Form/Nph/NphSampleFinalizeType.php
+++ b/src/Form/Nph/NphSampleFinalizeType.php
@@ -64,6 +64,13 @@ class NphSampleFinalizeType extends NphOrderForm
                                 if ($condition && empty($value)) {
                                     $context->buildViolation('Barcode is required')->addViolation();
                                 }
+                            }),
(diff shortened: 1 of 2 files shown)


shyamzcode (changes requested):
@Ehaic You might need to handle the edit aliquot logic as well. Currently, while editing it's throws me the error message even though it has one aliquot that is already entered.

image

shyamzcode:
> > @Ehaic You might need to handle the edit aliquot logic as well. Currently, while editing it's throws me the error message even though it has one aliquot that is already entered.
> > image
> 
> But in your image the 500 one that already exists is under 500uL so isn't the message still correct? I just tried editing a sample that had an existing 500uL sample and it went through fine.

Edit is working now

PR #1455, opened 2023-05-25
[HPRO-1419] Fixed feedback, display timezones, remove participant info from banner.

Ehaic:
| Q                   | A
| Target Release?     | 3.7.0 
| Bug fix?            | ✅
| New feature?        | ❌ 
| Database migration? | ❌ 
| New configuration?  | ❌ 
| Composer updates?   | ❌ 
| NPM updates?        | ❌ 
| Jira Ticket(s)      | HPRO-1419

### DIFF
--- a/templates/program/nph/order/sample-finalize.html.twig
+++ b/templates/program/nph/order/sample-finalize.html.twig
@@ -76,18 +76,18 @@
         <div class="col-sm-7">
             {% if sample.rdrId and sample.finalizedTs %}
                 <div class="alert alert-success well-sm">
-                    {% if sample.biobankFinalized %}
+                    {% if biobankView %}
                         <i class="fa fa-exclamation-circle"></i>
                         <strong class="warning-text">
                             Finalized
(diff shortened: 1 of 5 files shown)


templates/program/nph/order/sample-finalize.html.twig
```diff
@@ -72,7 +81,7 @@
                         <strong class="warning-text">
                             Finalized
                             by Biobank user
-                            on {{ sample.finalizedTs|date('n/j/Y g:ia', app.user.timezone) }}
+                            on {{ macros.displayDateAndTimezone(sample.finalizedTs, sample.finalizedTimezoneId) }}
```
shyamzcode, line 84: I see that for biobank view you are displaying timezone only if the order is finalized by biobank user. I think we need to display timezone for other orders too if it's biobank view.

shyamzcode (changes requested):

shyamzcode (approved):

PR #1460, opened 2023-06-01
[HPRO-1137] Health Data Stream Sharing.

Ehaic:
| Q                   | A
| Target Release?     | 3.8.0 
| Bug fix?            | ❌ 
| New feature?        | ✅ 
| Database migration? | ❌ 
| New configuration?  | ❌ 
| Composer updates?   | ❌ 
| NPM updates?        | ❌ 
| Jira Ticket(s)      | HPRO-1137

### DIFF
--- a/src/Helper/WorkQueue.php
+++ b/src/Helper/WorkQueue.php
@@ -2693,7 +2693,7 @@ class WorkQueue
         return $headers;
     }
 
-    public static function mapExportColumns(&$columns): void
+    public static function mapExportColumns(& $columns): void
     {
         foreach ($columns as $column) {
             if (isset(self::$tableExportMap[$column])) {
@@ -2767,9 +2767,8 @@ class WorkQueue
         } else {
             if (!is_null($healthDataSharingStatus)) {
                 return self::dateFromString($healthDataSharingStatus, $userTimezone);
-            } else {
-                return '';
             }
+            return '';
         }
     }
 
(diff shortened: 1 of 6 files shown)


src/Helper/WorkQueue.php
```diff
@@ -1414,6 +1433,7 @@ class WorkQueue
         'retentionType',
         'isEhrDataAvailable',
         'latestEhrReceiptTime',
+        'healthDataStream',
```
shyamzcode, line 1436: We should always add new columns to the last in CSV export. Could you move this field to the last position?

shyamzcode (changes requested):

src/Helper/WorkQueue.php
```diff
@@ -2609,6 +2632,19 @@ public static function getRetentionType($value)
         }
     }
 
+    public static function getHealthDataSharingStatus($value, $time, $userTimezone)
```
shyamzcode, line 2635: Could you add type hints for these params and return type for the method. Also could you write unit tests for these public methods?

src/Helper/WorkQueue.php
```diff
@@ -2715,6 +2751,21 @@ public static function csvDigitalHealthSharingStatus($digitalHealthSharingStatus
         return !$displayDate ? 0 : '';
     }
 
+    public static function csvHealthDataSharingStatus($healthDataSharingStatus, $displayDate = false, $userTimezone = null)
```
shyamzcode, line 2754: As as mentioned above could you type hint and write unit tests for this method?

shyamzcode (changes requested):

PR #1467, opened 2023-06-12
Display Ancillary Studies

Ehaic:
| Q                   | A
| Target Release?     | 3.9.0 
| Bug fix?            | ❌ 
| New feature?        | ✅ 
| Database migration? | ❌ 
| New configuration?  | ❌ 
| Composer updates?   | ❌ 
| NPM updates?        | ❌ 
| Jira Ticket(s)      | HPRO-... 

### Screenshots 
image
image

### DIFF
--- a/src/Helper/WorkQueue.php
+++ b/src/Helper/WorkQueue.php
@@ -1236,7 +1236,7 @@ class WorkQueue
             'rdrField' => 'reconsentForStudyEnrollmentAuthored',
             'csvFormatDate' => true
         ],
-        'consentForNphModule1' => [
+        'NphStudyStatus' => [
             'name' => 'NPH Module 1 Consent',
             'group' => 'ancillaryStudies',
             'rdrField' => 'genderIdentity',
@@ -1335,7 +1335,7 @@ class WorkQueue
         'genderIdentity',
         'race',
         'education',
-        'consentForNphModule1'
+        'NphStudyStatus'
     ];
 
     public static $columnGroups = [
@@ -1585,7 +1585,7 @@ class WorkQueue
         'genderIdentity',
         'race',
         'education',
-        'consentForNphModule1'
+        'NphStudyStatus'
     ];
 
     public static $consentSortColumns = [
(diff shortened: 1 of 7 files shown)


src/Helper/Participant.php
```diff
@@ -408,6 +414,30 @@ private function parseRdrParticipant($participant)
         if (isset($participant->aian)) {
             $this->isAIAN = $participant->aian;
         }
+
+        if (isset($participant->nphWithdrawal)) {
+            $this->nphWithdrawal = $participant->nphWithdrawal;
```
shyamzcode, line 419: What is the intention behind setting these properties? Usually, we set properties of the Participant class when there is a computation involved with the RDR participant object. But here since you are just saving the property directly not sure what's the purpose behind it.

src/Helper/WorkQueue.php
```diff
@@ -2936,4 +2970,25 @@ public static function displayDateStatus($time, $userTimezone, $displayTime = tr
         }
         return self::HTML_DANGER;
     }
+
+    public static function getNphStudyStatus(Participant $participant, string $userTimezone, bool $displayTime = false): string
```
shyamzcode, line 2974: Could you write unit tests for this public method?

src/Helper/WorkQueue.php
```diff
@@ -2936,4 +2970,25 @@ public static function displayDateStatus($time, $userTimezone, $displayTime = tr
...
+            return self::HTML_DANGER . ' ' . self::dateFromString($participant->nphWithdrawalAuthored, $userTimezone, $displayTime);
+        } elseif ($participant->nphDeactivation) {
+            return self::HTML_DANGER . ' ' . self::dateFromString($participant->nphDeactivationAuthored, $userTimezone, $displayTime);
+        } elseif ($participant->consentForNphModule1) {
+            return self::HTML_SUCCESS . ' ' . self::dateFromString($participant->consentForNphModule1Authored, $userTimezone, $displayTime);
+        }
+        return self::HTML_DANGER;
+    }
+
+    public static function getCsvNphStudyStatus(Participant $participant, string $fieldKey, string $userTimezone): int|string
```
shyamzcode, line 2986: Same as above could you write unit tests for this public method?

src/Helper/Participant.php
shyamzcode: I believe PHPStan throws that error specifically when attempting to access the property in the Participant class. Are you experiencing any PHPStan errors when you remove those setters? Considering that you are only setting those properties in the Participant class and not accessing them, I think it's safe to remove them.

PR #1617, opened 2023-11-10
[HPRO-1704] Readd checkmark for GROR Consent.

Ehaic:
| Q                   | A
| Bug fix?            | ✅ 
| New feature?        | ❌ 
| Database migration? | ❌ 
| New configuration?  | ❌ 
| Composer updates?   | ❌ 
| NPM updates?        | ❌ 
| Jira Ticket(s)      | HPRO-1704 

### Screenshots 
image
image

### DIFF
--- a/templates/macros/display-text.html.twig
+++ b/templates/macros/display-text.html.twig
@@ -126,15 +126,24 @@
     {% endif %}
 {% endmacro %}
 
-{% macro displayConsentPanelStatus(participant, field, showConsentPDFs = '', showCheck = false) %}
+{% macro displayConsentPanelStatus(participant, field, showConsentPDFs = '') %}
     {% set consentValues = ['SUBMITTED', 'SUBMITTED_NO_CONSENT', 'SUBMITTED_NOT_SURE', 'SUBMITTED_INVALID'] %}
     {% if attribute(participant, field) in consentValues %}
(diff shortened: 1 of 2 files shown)


templates/macros/display-text.html.twig
```diff
@@ -126,10 +126,11 @@
     {% endif %}
 {% endmacro %}
 
-{% macro displayConsentPanelStatus(participant, field, showConsentPDFs = '') %}
+{% macro displayConsentPanelStatus(participant, field, showConsentPDFs = '', showCheck = false) %}
     {% set consentValues = ['SUBMITTED', 'SUBMITTED_NO_CONSENT', 'SUBMITTED_NOT_SURE', 'SUBMITTED_INVALID'] %}
     {% if attribute(participant, field) in consentValues %}
         {% set time = attribute(participant, field ~ 'Authored') %}
+        {% if showCheck %}<i class="fa fa-check text-success" aria-hidden="true"></i>{% endif %}
```
shyamzcode, line 133: I think the green tick needs to be displayed only when the value is `SUBMITTED`. Might need to check with Gwen and confirm.

shyamzcode (changes requested):

templates/macros/display-text.html.twig
Ehaic: Setup a new function to match the workqueue logic for the consent.

PR #1685, opened 2023-12-28
[HPRO-1532][HPRO-1769] NPH Workqueue columns and column visibility

Ehaic:
| Q                   | A
| Bug fix?            | ❌ 
| New feature?        | ✅ 
| Database migration? | ❌ 
| New configuration?  | ❌ 
| Composer updates?   | ❌ 
| NPM updates?        | ✅ 
| Jira Ticket(s)      | HPRO-1532, HPRO-1769


src/WorkQueue/ColumnDefs/NPH/config.json
```diff
@@ -0,0 +1,172 @@
...
+      "dataField": "nphModule3ConsentStatus",
+      "class": "moduleConsentStatus",
+      "group": "Consents",
+      "enable": true
+    },
+    {
+      "display": "Paired Site",
+      "sortable": false,
+      "filterable": false,
+      "dataField": "nphPairedSite",
```
shyamzcode, line 119: We might need to display the site name instead of the site ID. You could use the getNphSiteDisplayName method in the SiteServe class to implement this.

src/WorkQueue/ColumnDefs/NPH/moduleConsentStatus.php
```diff
@@ -0,0 +1,26 @@
+<?php
+
+namespace App\WorkQueue\ColumnDefs\NPH;
+
+use App\WorkQueue\ColumnDefs\defaultColumn;
+
+class moduleConsentStatus extends defaultColumn
+{
+    public function getColumnDisplay($data, $dataRow): string
```
shyamzcode, line 9: Could you add return types for these parameters and in other places as well?

src/WorkQueue/DataSources/NphDataSource.php
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    public function getWorkqueueData(int $offset, int $limit, ColumnCollection $columnCollection): array
+    {
+        $response = $this->rdrApi->GQLPost('rdr/v1/nph_participant', $this->getSearchQuery($offset, $limit));
+        $result = json_decode($response->getBody()->getContents(), true);
+        return $result;
+    }
+
+    private function getSearchQuery($offset, $limit): string
```
shyamzcode, line 27: Could you add return types for these params?

shyamzcode (changes requested):
Overall, looks good! Could you add borders to the table columns and remove the default datatable search?

src/WorkQueue/ColumnDefs/NPH/moduleConsentStatus.php
Ehaic: I don't know of a good way to type these because the interface would have to be typed as well. The interface would need to be able to handle any sort of layout of the data (array, object, string, int, etc) as well as the datarow. I think the return type is enough since the other code doesn't care what its passing into these objects and only cares that strings are returned out of it. It's up to each individual implementation of a column to sort through its data and get what is trying to be displayed.

shyamzcode: Gotcha!

shyamzcode (approved):

PR #1696, opened 2024-01-16
[HPRO-1745][HPRO-1647] Add new enrollment status and updated to V3.2 enrollment status.

Ehaic:
| Q                   | A
| Bug fix?            | ❌ 
| New feature?        | ✅ 
| Database migration? | ❌ 
| New configuration?  | ❌ 
| Composer updates?   | ❌ 
| NPM updates?        | ❌ 
| Jira Ticket(s)      | HPRO-1745, HPRO-1647

### DIFF
--- a/src/Helper/Participant.php
+++ b/src/Helper/Participant.php
@@ -445,12 +445,12 @@ class Participant
         }
 
         // Determine core participant
-        if (!empty($participant->enrollmentStatus) && $participant->enrollmentStatus === 'FULL_PARTICIPANT') {
+        if (!empty($participant->enrollmentStatusV3_2) && $participant->enrollmentStatusV3_2 === 'FULL_PARTICIPANT') {
             $this->isCoreParticipant = true;
         }
 
         // Determine core minus pm participant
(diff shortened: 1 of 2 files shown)


shyamzcode (changes requested):
The sort columns are pulled from static sortColumns array (https://github.com/all-of-us/healthpro/blob/develop/src/Helper/WorkQueue.php#L1617). You might need to update that as well

Ehaic:
> The sort columns are pulled from static sortColumns array (https://github.com/all-of-us/healthpro/blob/develop/src/Helper/WorkQueue.php#L1617). You might need to update that as well

I believe that points back to the enrollmentStatus column which is already updated. I'll be sure and double check this once the updates are on stable.

shyamzcode:
@Ehaic The $sortColumns array needs to be updated as well, as that's the one we are sending to RDR when a field is selected for sorting. So, `enrollmentStatus` needs to be changed to `enrollmentStatusV3_2` in the $sortColumns array. You can check that by selecting the sort and logging which values we are sending to RDR for the sort field.

shyamzcode (changes requested):

shyamzcode (approved):

PR #1724, opened 2024-02-20
[HPRO-1717] Created Bulk Cancelling of samples.

Ehaic:
| Q                   | A
| Target Release?     | 4.3.5 
| Bug fix?            | ❌ 
| New feature?        | ✅ 
| Database migration? | ❌ 
| New configuration?  | ❌ 
| Composer updates?   | ❌ 
| NPM updates?        | ❌ 
| Jira Ticket(s)      | HPRO-1717

### DIFF
--- a/src/Service/Nph/NphOrderService.php
+++ b/src/Service/Nph/NphOrderService.php
@@ -601,22 +601,24 @@ class NphOrderService
     {
         $status = true;
         foreach ($formData as $sampleId => $checked) {
-            $sample = $this->em->getRepository(NphSample::class)->findOneBy(['sampleId' => $sampleId]);
-            if ($checked) {
-                $sampleObject = $this->getCancelRestoreRdrObject($type, $formData['reason']);
-                if ($sample->getRdrId()) {
-                    if ($this->cancelRestoreSample(
-                        $sample->getRdrId(),
-                        $sample->getNphOrder()->getParticipantId(),
-                        $type,
-                        $sampleObject
-                    )) {
-                        $this->saveSampleModificationsData($sample, $type, $formData);
+            if (isset($checked) && $checked !== null && is_bool($checked) && $checked) {
+                $sample = $this->em->getRepository(NphSample::class)->findOneBy(['sampleId' => $sampleId]);
+                if ($sample !== null) {
+                    $sampleObject = $this->getCancelRestoreRdrObject($type, $formData['reason']);
+                    if ($sample->getRdrId()) {
+                        if ($this->cancelRestoreSample(
+                            $sample->getRdrId(),
+                            $sample->getNphOrder()->getParticipantId(),
+                            $type,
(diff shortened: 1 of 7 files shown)


src/Service/Nph/NphOrderService.php
```diff
@@ -597,6 +597,32 @@ public function saveSamplesModification(array $formData, string $type, NphOrder
...
     }
 
+    public function updateSampleModificationBulk(array $formData, string $type): bool
+    {
+        $status = true;
+        foreach ($formData as $sampleId => $checked) {
+            $sample = $this->em->getRepository(NphSample::class)->findOneBy(['sampleId' => $sampleId]);
+            if ($checked) {
+                $sampleObject = $this->getCancelRestoreRdrObject($type, $formData['reason']);
+                if ($sample->getRdrId()) {
```
shyamzcode, line 607: RDR id is only available for samples that are sent to RDR. You might need to handle the samples that do not have the RDR id set.

shyamzcode (changes requested):
Looks like you missed adding the "Cancel All" button in the Participant summary page
image

src/Service/Nph/NphOrderService.php
```diff
@@ -597,6 +597,32 @@ public function saveSamplesModification(array $formData, string $type, NphOrder
         return $status;
     }
 
+    public function updateSampleModificationBulk(array $formData, string $type): bool
+    {
+        $status = true;
+        foreach ($formData as $sampleId => $checked) {
```
shyamzcode, line 603: You are iterating through form data which will contain other data key-value pairs like "reason," "otherText," etc., which you want to exclude. That's why it's throwing the error below.

image

src/Service/Nph/NphOrderService.php
```diff
@@ -597,6 +597,32 @@ public function saveSamplesModification(array $formData, string $type, NphOrder
...
+            $sample = $this->em->getRepository(NphSample::class)->findOneBy(['sampleId' => $sampleId]);
```
shyamzcode, line 604: Could you run this only when the sample is checked? This will prevent unnecessary db queries.

src/Service/Nph/NphOrderService.php
```diff
@@ -597,6 +597,34 @@ public function saveSamplesModification(array $formData, string $type, NphOrder
         return $status;
     }
 
+    public function updateSampleModificationBulk(array $formData, string $type): bool
+    {
+        $status = true;
+        foreach ($formData as $sampleId => $checked) {
+            if (isset($checked) && $checked !== null && is_bool($checked) && $checked) {
```
shyamzcode, line 604: This could be simplified to just $checked === true right?

Ehaic: Yes you are correct.

shyamzcode (approved):

PR #1762, opened 2024-03-25
[HPRO-1630] Save user supplied alternative measurements in physical evaluations.

Ehaic:
| Q                   | A
| Bug fix?            | ❌ 
| New feature?        | ✅ 
| Database migration? | ✅ 
| New configuration?  | ❌ 
| Composer updates?   | ❌ 
| NPM updates?        | ❌ 
| Jira Ticket(s)      | HPRO-1630


shyamzcode:
@Ehaic I am encountering this error when trying to create new physical measurements. Also, I noticed the same behavior when opening an existing PM form.

image

Ehaic:
@shyamzcode Should be fixed now, I missed the replication field part for the alternate units.

web/assets/js/views/PhysicalEvaluation-0.3.js
```diff
@@ -615,6 +615,15 @@ PMI.views["PhysicalEvaluation-0.3"] = Backbone.View.extend({
                 inches += parseFloat($("#alt-units-height-in").val());
             }
             val = this.inToCm(inches);
+        } else if (type == "alt-units-height-ftin") {
```
shyamzcode, line 618: It seems like you are trying to display the user-entered values in the top right corner. Should we apply this same approach to other fields such as weight, waist circumference, and hip circumference? Also, this entire process needs to be handled for pediatric forms as well.

shyamzcode (changes requested):

web/assets/js/views/PhysicalEvaluation-0.3.js
Ehaic: Actually that part of the code is for updating the original entry (height cm in this case) using the converted value of the alternative units. I had to update this particular part just for the height code. The other stuff still works fine. You are right that I need to update the top right display though. I'll fix that.

shyamzcode (approved):

PR #1767, opened 2024-03-28
[HPRO-1847] Set incentive recipient to the related participant if guardian recipient

Ehaic:
| Q                   | A
| Bug fix?            | ✅ 
| New feature?        | ❌ 
| Database migration? | ❌ 
| New configuration?  | ❌ 
| Composer updates?   | ❌ 
| NPM updates?        | ❌ 
| Jira Ticket(s)      | HPRO-1847

### DIFF
--- a/src/Controller/ParticipantDetailsController.php
+++ b/src/Controller/ParticipantDetailsController.php
@@ -308,7 +308,6 @@ class ParticipantDetailsController extends BaseController
         if ($incentive && $incentive->getRecipient() == Incentive::PEDIATRIC_GUARDIAN) {
             $incentive->setRecipient($incentive->getRelatedParticipantRecipient());
         }
-        $incentive->setRecipient($incentive->getRelatedParticipantRecipient());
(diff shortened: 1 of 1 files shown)


src/Controller/ParticipantDetailsController.php
```diff
@@ -305,6 +305,10 @@ public function participantIncentiveAction(
         if ($incentive && $incentive->getSite() !== $siteService->getSiteId()) {
             throw $this->createAccessDeniedException();
         }
+        if ($incentive && $incentive->getRecipient() == Incentive::PEDIATRIC_GUARDIAN) {
+            $incentive->setRecipient($incentive->getRelatedParticipantRecipient());
+        }
+        $incentive->setRecipient($incentive->getRelatedParticipantRecipient());
```
shyamzcode, line 311: If the intention is to set the recipient of $incentive to the related participant recipient only when the if condition is met, then the second setRecipient call should probably be removed.

shyamzcode (changes requested):

shyamzcode (approved):

PR #1794, opened 2024-04-24
[HPRO-1560] Only send emails if staging or prod

Ehaic:
| Q                   | A
| Bug fix?            | ❌ 
| New feature?        | ✅ 
| Database migration? | ❌ 
| New configuration?  | ❌ 
| Composer updates?   | ❌ 
| NPM updates?        | ❌ 
| Jira Ticket(s)      | HPRO-1560

### DIFF
--- a/src/Service/EmailNotificationService.php
+++ b/src/Service/EmailNotificationService.php
@@ -10,7 +10,7 @@ abstract class EmailNotificationService
     protected $managerRegistry;
     protected $participantSummaryService;
     protected $loggerService;
-    protected $env;
+    protected EnvironmentService $env;
     protected $params;
     protected $twig;
     protected $siteRepository;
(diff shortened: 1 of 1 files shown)


src/Service/EmailNotificationService.php
```diff
@@ -61,14 +61,16 @@ public function sendEmails()
...
                 } else {
-                    $message = new Message($this->env, $this->loggerService, $this->twig, $this->params);
-                    $message
-                        ->setTo($result['emails'])
-                        ->render($this->render, [
-                            $this->level => $result['id'],
-                            'participants' => $participants
-                        ])
-                        ->send();
+                    if ($this->env->isStaging() || $this->env->isProd()) {
```
shyamzcode, line 64: We should should send emails in stable and prod and disable in dev and staging env.

src/Service/EmailNotificationService.php
```diff
@@ -61,14 +61,16 @@ public function sendEmails()
...
                 } else {
-                    $message = new Message($this->env, $this->loggerService, $this->twig, $this->params);
-                    $message
-                        ->setTo($result['emails'])
-                        ->render($this->render, [
-                            $this->level => $result['id'],
-                            'participants' => $participants
-                        ])
-                        ->send();
+                    if ($this->env->isStable() || $this->env->isProd()) {
```
shyamzcode, line 64: This won't log the message in the logs as the send method is not called. I think the best way to do this is to remove the mail_method key from configurations, which will then prevent sending emails but still log.

Below is the code where the mail method is checked.
https://github.com/all-of-us/healthpro/blob/develop/src/Service/Message.php#L81

shyamzcode (changes requested):

src/Service/EmailNotificationService.php
Ehaic: You're right. Good call I didn't notice that.

PR #1834, opened 2024-06-26
[HPRO-1932] DV Site Config Option for enabling PST Tubes

Ehaic:
| Q                   | A
| Bug fix?            | ❌ 
| New feature?        | ✅ 
| Database migration? | ❌ 
| New configuration?  | ❌ 
| Composer updates?   | ❌ 
| NPM updates?        | ❌ 
| Jira Ticket(s)      | HPRO-1932

### DIFF
--- a/src/Controller/OrderController.php
+++ b/src/Controller/OrderController.php
@@ -175,7 +175,7 @@ class OrderController extends BaseController
                 if ($session->get('siteType') === 'dv' && $this->siteService->isDiversionPouchSite()) {
                     $order->setType('diversion');
                 }
-                if ($session->get('siteType') === 'dv' && $this->siteService->isDiversionPouchSite() === false && $params->get('order_samples_version_dv') > 3.1) {
+                if ($session->get('siteType') === 'dv' && $this->siteService->isDiversionPouchSite() === false && (double) $params->get('order_samples_version_dv') > 3.1) {
                     $order->setVersion(null);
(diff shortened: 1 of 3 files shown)


src/Controller/OrderController.php
```diff
@@ -175,7 +175,8 @@ public function orderCreateAction($participantId, Request $request, SessionInter
                 if ($session->get('siteType') === 'dv' && $this->siteService->isDiversionPouchSite()) {
                     $order->setType('diversion');
                 }
-                if ($session->get('siteType') === 'dv' && $this->siteService->isDiversionPouchSite() === false && $order->getCurrentVersion() > 3.1 && $order->getVersion() === null) {
+                if ($session->get('siteType') === 'dv' && $this->siteService->isDiversionPouchSite() === false && (float) $params->get('order_samples_version_dv') > 3.1) {
```
shyamzcode, line 178: We might want to set a default version when `order_samples_version_dv` config variable is not defined else it will throw an error.

shyamzcode (changes requested):
One thing I noticed is this works fine if we set the version to:
order_samples_version: 3.1
order_samples_version_dv: 3.2

But if we set the versions to:
order_samples_version: 3.2
order_samples_version_dv: 3.1

I am getting this error in the DV workflow, which might be due to the version being less than the actual order version. You might need to handle this scenario as well.

image

shyamzcode (approved):

PR #1892, opened 2024-09-26
[HPRO-2022] Add code to create and update DLW's directly with RDR

Ehaic:
| Q                   | A
| Bug fix?            | ❌ 
| New feature?        | ✅ 
| Database migration? | ✅ 
| New configuration?  | ❌ 
| Composer updates?   | ❌ 
| NPM updates?        | ❌ 
| Jira Ticket(s)      | HPRO-2022


src/Entity/NphDlw.php
```diff
@@ -48,6 +48,9 @@ class NphDlw
     #[ORM\Column(length: 50, nullable: true)]
     private ?string $visitPeriod = null;
 
+    #[ORM\Column(type: Types::TEXT, nullable: true)]
+    private ?string $rdr_id = null;
```
shyamzcode, line 52: Could you use camel case (rdrId) instead of an underscore? You might need to update all the getters, setters, and other related places as well.

migrations/Version20240926151434.php
```diff
@@ -0,0 +1,35 @@
...
+
+declare(strict_types=1);
+
+namespace DoctrineMigrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
```
shyamzcode, line 11: Could you remove these auto generated comments here and in the below methods as well?

shyamzcode (changes requested):

src/Entity/NphDlw.php
shyamzcode: The database field can remain rdr_id, but in the entity, you can name it rdrId.

src/Service/Nph/NphDlwBackfillService.php
```diff
@@ -0,0 +1,29 @@
...
+class NphDlwBackfillService
+{
+    protected EntityManagerInterface $em;
+    protected NphDlwRepository $NphDlwRepository;
+    protected NphOrderService $nphOrderService;
+
+    public function __construct(EntityManagerInterface $em, NphDlwRepository $dlwRepository, NphOrderService $nphOrderService)
+    {
+        $this->em = $em;
+        $this->NphDlwRepository = $dlwRepository;
```
shyamzcode, line 18: @Ehaic I think you can remove this as you are directly calling the repository using $this->em->getRepository on line 24

shyamzcode (approved):

